PR #37, opened 2021-12-16
Add metrics package

liamg:


rules/register.go
```diff
@@ -15,7 +18,9 @@ func (r RegisteredRule) Evaluate(s *state.State) Results {
 	if r.checkFunc == nil {
 		return nil
 	}
+	ruleTimer := metrics.Timer("defsec-rule-timers", r.Rule().LongID(), true)
```
owenrumney, line 21: should this only be done if debug set to true? then we can save the overhead of doing something nobody will see?

liamg: yeah, good point - we'll have to pass into defsec that we're in `debug` somehow - i'll remove the metric for now and we can figure that out in another PR if that works for you?

owenrumney (approved):

PR #81, opened 2022-01-16
Add check for Digital Ocean's Kubernetes auto_upgrade option.

atombrella:
Again, I'm not really sure about tests, and if `tfsec` is just the place to add this stuff. It feels a bit weird adding code without tests :man_shrugging:


owenrumney (approved):
:+1:

PR #255, opened 2022-02-09
chore: migrate adapters into defsec, relocate types to trivy-config-parsers

liamg:


owenrumney (approved):
I read all 1,092 changes

PR #256, opened 2022-02-10
fix: involve the public access block in the check for public acl

owenrumney:


provider/aws/s3/bucket.go
```diff
@@ -16,6 +16,10 @@ type Bucket struct {
 func (b *Bucket) HasPublicExposureACL() bool {
 	for _, publicACL := range []string{"public-read", "public-read-write", "website", "authenticated-read"} {
 		if b.ACL.EqualTo(publicACL) {
+			// if there is a public access block, check the public ACL blocks
+			if b.PublicAccessBlock.IsManaged() {
```
liamg, line 20: Can PublicAccessBlock be `<nil>` here? Can't remember if it's a pointer or not.

owenrumney: the broken tests suggest yes :grimacing:

owenrumney: Fixed @liamg

liamg (approved):

PR #273, opened 2022-02-17
feat: Add iam line numbers

liamg:
We now support line number calculation from parsing of IAM policies. This is best supported by using literal strings or by defining IAM policies using HCL syntax.


owenrumney (approved):
LGTM

PR #286, opened 2022-02-23
chore: Add support for automatic PRs from dependabot

atombrella:
The file is just copied from tfsec. The `github-actions` and `docker` section are removed, as I don't think they're relevant in this repository.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+      - dependency-name: "github.com/zclconf/go-cty"
+        versions: ["1.x"]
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  # Maintain dependencies for go-modules
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
owenrumney, line 12: this seems to be missing the important ignore to stop go-cty being updated automatically. https://github.com/aquasecurity/tfsec/blob/master/.github/dependabot.yml#L13-L15

We have to stay in sync with HCL library version

owenrumney (approved):

PR #307, opened 2022-03-02
feat: add AWS provider v4 support

owenrumney:
Add support for V4 changes

### DIFF
--- a/adapters/terraform/aws/s3/bucket.go
+++ b/adapters/terraform/aws/s3/bucket.go
@@ -186,7 +186,7 @@ func isVersioned(b *terraform.Block) types.BoolValue {
 		if status.Equals("Enabled", terraform.IgnoreCase) {
 			return types.Bool(true, status.GetMetadata())
 		} else {
-			return types.Bool(false, versioningBlock.GetMetadata())
+			return types.Bool(false, b.GetMetadata())
 		}
 	}
 	return types.BoolDefault(
--- a/adapters/terraform/aws/s3/bucket.go
+++ b/adapters/terraform/aws/s3/bucket.go
@@ -184,7 +184,7 @@ func isVersioned(b *terraform.Block) types.BoolValue {
 	if versioningBlock := b.GetBlock("versioning_configuration"); versioningBlock.IsNotNil() {
 		status := versioningBlock.GetAttribute("status")
 		if status.Equals("Enabled", terraform.IgnoreCase) {
-			return types.Bool(true, versioningBlock.GetMetadata())
+			return types.Bool(true, status.GetMetadata())
 		} else {
 			return types.Bool(false, versioningBlock.GetMetadata())
 		}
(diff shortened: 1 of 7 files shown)


adapters/terraform/aws/s3/bucket.go
```diff
@@ -74,13 +171,24 @@ func hasLogging(b *terraform.Block) types.BoolValue {
...
 }
 
 func isVersioned(b *terraform.Block) types.BoolValue {
 	if versioningBlock := b.GetBlock("versioning"); versioningBlock.IsNotNil() {
 		return versioningBlock.GetAttribute("enabled").AsBoolValueOrDefault(true, versioningBlock)
 	}
+	if versioningBlock := b.GetBlock("versioning_configuration"); versioningBlock.IsNotNil() {
+		status := versioningBlock.GetAttribute("status")
+		if status.Equals("Enabled", terraform.IgnoreCase) {
+			return types.Bool(true, versioningBlock.GetMetadata())
```
liamg, line 187:
```suggestion
			return types.Bool(true, status.GetMetadata())
```

adapters/terraform/aws/s3/bucket.go
```diff
@@ -74,13 +171,24 @@ func hasLogging(b *terraform.Block) types.BoolValue {
 		}
 		return types.BoolDefault(false, loggingBlock.GetMetadata())
 	}
+	if targetBucket := b.GetAttribute("target_bucket"); targetBucket.IsNotNil() {
+		return types.Bool(true, b.GetMetadata())
```
liamg, line 175:
```suggestion
		return types.Bool(true, targetBucket.GetMetadata())
```

adapters/terraform/aws/s3/bucket.go
```diff
@@ -74,13 +171,24 @@ func hasLogging(b *terraform.Block) types.BoolValue {
...
 func isVersioned(b *terraform.Block) types.BoolValue {
 	if versioningBlock := b.GetBlock("versioning"); versioningBlock.IsNotNil() {
 		return versioningBlock.GetAttribute("enabled").AsBoolValueOrDefault(true, versioningBlock)
 	}
+	if versioningBlock := b.GetBlock("versioning_configuration"); versioningBlock.IsNotNil() {
+		status := versioningBlock.GetAttribute("status")
+		if status.Equals("Enabled", terraform.IgnoreCase) {
+			return types.Bool(true, versioningBlock.GetMetadata())
+		} else {
+			return types.Bool(false, versioningBlock.GetMetadata())
```
liamg, line 189: The block will be `nil` here so we need the parent blocks metadata.
```suggestion
			return types.Bool(false, b.GetMetadata())
```

liamg (changes requested):
Couple of metadata swaps, otherwise all good

liamg (approved):

PR #341, opened 2022-03-08
feat: Split checks for performance insight enable and encrypt out

### LINKED ISSUE #340: bug: checks/aws/rds/enable-performance-insights
**Describe the bug**
Encryption for Performance Insights seems to be required even though Performance Insights is disabled.
**Note:** The [tfsec documentation](https://aquasecurity.github.io/tfsec/v1.6.2/checks/aws/rds/enable-performance-insights/) for this issue seems to be wrong, or at least very confusing: The URL says "enable-performance-insights", the title says "Encryption for RDS Performance Insights should be enabled", explanation says "When enabling Performance Insights on an RDS cluster or RDS DB Instance, and encryption key should be provided", and the tool itself reports "Instance does not have performance insights enabled".

I don't understand how not having Performance Insights enabled is a security risk. If the encryption is the problem, then it shouldn't report the error if Performance Insights is disabled. And especially when encryption actually is enabled.

**To Reproduce**
1. Run the Github workflow in https://github.com/MartinSMadsen/terraform-security-scan.

**Expected behavior**
Tool does not report an error for Performance Insights if Performance Insights is disabled.

**Screenshots/Output**

From https://github.com/MartinSMadsen/terraform-security-scan/runs/5467333705:

image

**System Info**

 - tfsec version: `master` (using https://github.com/triat/terraform-security-scan)
 - terraform version: `?`
 - OS: `ubuntu-latest`

**Example Code**
See https://github.com/MartinSMadsen/terraform-security-scan.

owenrumney:
Resolves #340

### DIFF
--- a/avd_docs/aws/rds/AVD-AWS-0133/CloudFormation.md
+++ b/avd_docs/aws/rds/AVD-AWS-0133/CloudFormation.md
@@ -1,0 +1,15 @@
+
+Enable Performance Insights to detect potential problems
+
+```yaml
+---
+AWSTemplateFormatVersion: "2010-09-09"
+Description: A sample template
+AWSTemplateFormatVersion: 2010-09-09
+Description: Good example
+Resources:
+  Queue:
+    Type: AWS::RDS::DBInstance
+    Properties:
+      EnablePerformanceInsights: true
+```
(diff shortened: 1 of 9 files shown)


liamg (approved):
:rocket:

PR #348, opened 2022-03-10
feat: Add Deregister rule function

liamg:


owenrumney (approved):
I was sure there was already a Deregister

PR #444, opened 2022-03-28
test: add tests for aws ssm tf adapters

### LINKED ISSUE #443: aws ssm - add tests for terraform adapters
See #284

vanesasejdiu:
resolves #443


vanesasejdiu:
additional changes: when kms key is a resource reference, grab the full name of the referenced block so it doesn't alert as empty

if there's a better approach to this, do let me know ¯\\\_(ツ)\_/¯

liamg (approved):

PR #449, opened 2022-03-28
fix: fix 448 classic elb doesn't have same attributes as alb

### LINKED ISSUE #448: bug: False positive for aws-elb-drop-invalid-headers
I've recently upgraded to TFSec 1.8.0. I have a classic ELB definition:

```
resource "aws_elb" "elb" {
```

and I'm now getting a TFSec alert for this resource: `Result aquasecurity/tfsec#1 HIGH Application load balancer is not set to drop invalid headers.`

The error message refers to "Application load balancer" but this is not an ALB, it's a [Classic] ELB.

From what I can see on https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elb there is no option for classic ELBs to drop invalid headers. So it's not clear if/how I can resolve this alert, other than to simply ignore it. Am I missing something or is this a false positive?

owenrumney:
Resolves #448 

Signed-off-by: Owen Rumney <owen@owenrumney.co.uk>


liamg (approved):

tom-dalton-fanduel:
Brilliant, thank you!

PR #457, opened 2022-03-29
Refactor for fs.FS based scans

liamg:
This refactors scanners to implement the interface:
```
type Scanner interface {
	ScanFS(ctx context.Context, fs fs.FS, dir string) (scan.Results, error)
}
```
where `fs.FS` refers to https://pkg.go.dev/io/fs#FS

It also introduces the `*universal.Scanner` which can scan a filesystem for all supported config types (terraform, cloudformation, dockerfile and kubernetes).


owenrumney (dismissed):
Whats the worst that could happen

owenrumney (approved):
My 350Mbps broadband took 10mins to open this PR

PR #485, opened 2022-04-06
fix(terraform): Fix path details for modules sourced from 'terraform init'

liamg:


liamg:
See aquasecurity/tfsec#1661

PR #493, opened 2022-04-07
feat(check): Add separate check for SNS encryption with CMK

### LINKED ISSUE #464: bug: aws-sns-enable-topic-encryption flags false positive if kms_master_key_id doesn't start with "/"

liamg:
Resolves #464


owenrumney (changes requested):
think this needs something added into avd_docs if its a new check

owenrumney (approved):

PR #501, opened 2022-04-07
feat(terraform): Add option to only apply rego policies

liamg:


owenrumney (commented):
Should this also allow for only running custom checks or do you think the should be deprecated in favour of rego?

liamg:
We could do that in tfsec I think, but defsec doesn't know about the concept of custom checks AFAIK

owenrumney (approved):

PR #507, opened 2022-04-10
Allow to skip checking required files

knqyf263:
## Description
There are 2 benefits:
- If the given fs.FS contains only target files, skipping `Required()` will bring better performance.
- In a way, it allows to override the behavior of `Required()`. The caller can decide which files are passed to `ScanFS()`.


liamg (approved):
Looks good, thanks :+1:

PR #519, opened 2022-04-20
fix(terraform): set module source in parsed ignores

tobiaszjankowski:
Fixes [1670 issue in tfsec](https://github.com/aquasecurity/tfsec/issues/1670).

The comparison between a parsed ignore entry and a detected issue is done by filenames reported by `Range` objects in the issue and the ignore. The filenames contain the module path (as `sourcePrefix`) and relative path to the source file. Since the detected issue's `Range` had an empty string as module path, the comparison always failed for downloaded modules (as local modules had an empty string as module source).


liamg (approved):
Awesome, thanks @tobiaszjankowski :rocket:

PR #521, opened 2022-04-21
chore: downgrade some severities

sm171190:
Changed some severities for some checks


owenrumney:
@sm171190  - could you provide something to support/justify the changes in the severities?

sm171190:
> e

@owenrumney We're making the same changes in the Aqua commercial product. So, want to keep appshield in line with those changes.

owenrumney (approved):

PR #522, opened 2022-04-21
feat: Add filetype detection package

liamg:


pkg/detection/detect.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		_, ok := contents["Resources"]
+		return ok
+	}
+
+	matchers[FileTypeDockerfile] = func(name string, _ io.Reader) bool {
+		const requiredFile = "Dockerfile"
+		base := filepath.Base(name)
+		ext := filepath.Ext(base)
+		if strings.TrimSuffix(base, ext) == requiredFile {
```
owenrumney, line 84: I wonder if this should just be a strings.HasPrefix(name, "Dockerfile") because I think we use `Dockerfile.CI` 

Not sure if there is a convention for this naming though

liamg: I think that will do it - it will trim the extension off (`.CI`) and then `Dockerfile == Dockerfile`. I did it like that so we don;t match stuff like `dockerfile.go`.

owenrumney (dismissed):

owenrumney (approved):

PR #525, opened 2022-04-21
fix(plan): update scanning to cover more areas

owenrumney:

### DIFF
--- a/pkg/scanners/terraformplan/scanner.go
+++ b/pkg/scanners/terraformplan/scanner.go
@@ -2,7 +2,6 @@ package terraformplan
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"os"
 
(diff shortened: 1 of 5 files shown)


pkg/scanners/terraformplan/scanner.go
```diff
@@ -54,6 +55,12 @@ func (s *Scanner) Scan(reader io.Reader) (scan.Results, error) {
 		return nil, err
 	}
 
-	scanner := terraformScanner.New()
+	content, err := planFS.ReadFile("main.tf")
+	if err != nil {
+		panic(err)
```
liamg, line 60:
```suggestion
		return nil, err
```

pkg/scanners/terraformplan/scanner.go
```diff
@@ -54,6 +55,12 @@ func (s *Scanner) Scan(reader io.Reader) (scan.Results, error) {
...
+	}
+	fmt.Printf("Content: %s\n", string(content))
```
liamg, line 62: I think this is debug?

liamg (changes requested):
Couple of small comments, otherwise AOK

pkg/scanners/terraformplan/scanner.go
owenrumney: yeah, my bad. Removed

liamg (approved):

PR #543, opened 2022-04-27
fix: Add missing policy fs override

liamg:


pkg/rego/load.go
```diff
@@ -84,6 +84,11 @@ func (s *Scanner) LoadPolicies(loadEmbedded bool, srcFS fs.FS, paths []string, r
 		s.policies = make(map[string]*ast.Module)
 	}
 
+	if s.policyFS != nil {
+		s.debug.Log("Overriding filesystem for policies!")
```
owenrumney, line 88: spurious exclamation mark

owenrumney (approved):

pkg/rego/load.go
liamg: Yeah I got a bit overexcited about the override.

PR #559, opened 2022-05-04
feat(scanning): Add Helm Scanning support

owenrumney:


pkg/scanners/helm/parser/parser.go
```diff
@@ -0,0 +1,227 @@
...
+			if err != nil {
+				return err
+			}
+			if err := p.ParseFS(ctx, tarFS, "."); err != nil {
+				return err
+			}
+			return nil
+		}
+
+		return p.addPaths(ctx, path)
```
liamg, line 98: I'm not sure it matters, but it looks like the same path will get added multiple times as ParseFS is called again each time a tar is added. Will it cause any problems having duplicate paths in the slice? It doesn't look like it to me apart from being slightly inefficient but I don't think it's going to be noticeable.

pkg/scanners/helm/parser/parser.go
```diff
@@ -0,0 +1,227 @@
...
+
+func (p *Parser) writeBuildFiles(tempFs string) error {
+	for _, path := range p.filepaths {
+		content, err := fs.ReadFile(p.workingFS, path)
+		if err != nil {
+			return err
+		}
+		workingPath := strings.TrimPrefix(path, p.rootPath)
+		workingPath = filepath.Join(tempFs, workingPath)
+		if err := os.MkdirAll(filepath.Dir(workingPath), os.ModePerm); err != nil {
```
liamg, line 212: Should this be writing to a temporary filesystem instead of using `os.*`? Otherwise it'll write files into a project dir unless I'm misunderstanding (which is likely.)

pkg/scanners/helm/parser/parser.go
```diff
@@ -0,0 +1,227 @@
...
+			return err
+		}
+		if err := os.WriteFile(workingPath, content, os.ModePerm); err != nil {
```
liamg, line 215: Same as above

pkg/scanners/helm/scanner.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	chartFiles, err := helmParser.RenderedChartFiles()
+	if err != nil {
+		return nil, err
+	}
+
+	var results []scan.Result
+	regoScanner := rego.NewScanner()
+	if err := regoScanner.LoadPolicies(true, nil, nil, nil); err != nil {
```
liamg, line 99: Need to pass these options through for loading rego policies etc.

pkg/scanners/helm/test/scanner_test.go
```diff
@@ -0,0 +1,138 @@
...
+			if _, exists := visited[id]; !exists {
+				visited[id] = true
+				errorCodes = append(errorCodes, id)
+			}
+		}
+		assert.Len(t, errorCodes, 19)
+
+		sort.Strings(errorCodes)
+
+		assert.Equal(t, []string{"AVD-KSV-0001", "AVD-KSV-0003", "AVD-KSV-0004",
```
liamg, line 63: You can use an option to specify your own single rego policy in testdata so you can always expect one result that never changes - can show you later on.

pkg/scanners/helm/test/scanner_test.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() { _ = out.Close() }()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	return out.Close()
```
liamg, line 137: There's already an `out.Close()` in a defer a few lines up, can remove one of these.

pkg/scanners/helm/test/testdata/expected/mysql/templates/secrets.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: mysql
+  namespace: 
+  labels:
+    app.kubernetes.io/name: mysql
+    helm.sh/chart: mysql-8.8.26
+    app.kubernetes.io/instance: mysql
+    app.kubernetes.io/managed-by: Helm
+type: Opaque
+data:
+  mysql-root-password: "aGZYYW1vN3V5NA=="
```
liamg, line 14: Squealer is going to have a great time here :D

pkg/scanners/kubernetes/parser/manifest.go
```diff
@@ -0,0 +1,28 @@
...
+
+	switch value.Tag {
+	case "!!map":
+		node := make(ManifestNode)
+		if err := value.Decode(&node); err != nil {
+			return err
+		}
+		m.Content = node
+	default:
+		panic(value.Tag)
```
liamg, line 20: Can we return an error here instead? Or is it game over if this is wrong?

pkg/scanners/helm/test/scanner_test.go
```diff
@@ -0,0 +1,138 @@
...
+		visited := make(map[string]bool)
+		var errorCodes []string
+		for _, result := range failed {
+			id := result.Flatten().RuleID
+			if _, exists := visited[id]; !exists {
+				visited[id] = true
+				errorCodes = append(errorCodes, id)
+			}
+		}
+		assert.Len(t, errorCodes, 19)
```
liamg, line 59: Would be cool to have a test that checks for start/end lines as it seems to be broken everywhere in the other scanners :(

liamg (changes requested):

pkg/scanners/helm/parser/parser.go
owenrumney: I don't see the flow where that happens. If the tar is added, it returns after its parsed it, if not, the path is added here.

pkg/scanners/helm/parser/parser.go
owenrumney: No, this is writing to the physical on disk temp location that will ultimately go into the helm library

pkg/scanners/helm/parser/parser.go
owenrumney: same as above

pkg/scanners/helm/scanner.go
owenrumney: I wanted to ask you about this....

pkg/scanners/helm/test/scanner_test.go
owenrumney: :+1:

pkg/scanners/helm/test/scanner_test.go
owenrumney: good point, I added the defer forgot to remove this

pkg/scanners/helm/test/testdata/expected/mysql/templates/secrets.yaml
owenrumney: Squealer Rocks People! :guitar:

pkg/scanners/kubernetes/parser/manifest.go
owenrumney: nope, its half finished with line numbers

pkg/scanners/helm/test/scanner_test.go
owenrumney: yeah, could add that when I start getting line numbers working

liamg (approved):

PR #565, opened 2022-05-05
feat: Adding line number support for k8s rego policies

### LINKED ISSUE #564: feat: Kubernetes misconfig line numbers
Add line numbers for the findings of kubernetes rego policies.

When scanning kubernetes manifests, we want to be able to narrow down the area of the issue.

owenrumney:
Resolves #564 

- Update the rego to support line numbers
- Parse Kubernetes manifest into Manifest Nodes to get line numbers
- Load filename in the result
- Fix links between Helm parsing and forwarding to K8s parser


pkg/scanners/universal/scanner.go
```diff
@@ -40,6 +41,7 @@ func New(opts ...options.ScannerOption) *Scanner {
 			json.NewScanner(opts...),
 			yaml.NewScanner(opts...),
 			toml.NewScanner(opts...),
+			helm.New("mysql", opts...),
```
liamg, line 44: Guessing this should be something else and not "mysql" but I'm not sure what.

liamg (changes requested):

pkg/scanners/universal/scanner.go
owenrumney: whoops

PR #574, opened 2022-05-09
fix: Avoid duplicate rule evaluation for rules with multiple definitions

liamg:


pkg/scanners/helm/parser/parser.go
```diff
@@ -232,7 +232,6 @@ func (p *Parser) writeBuildFiles(tempFs string) error {
 		if err != nil {
 			return err
 		}
-		fmt.Printf("path: %s, rootPath: %s\n", path, p.rootPath)
```
owenrumney, line 235: I see you're cleaning up after me.....

pkg/scanners/helm/test/scanner_test.go
```diff
@@ -47,7 +47,7 @@ func Test_helm_scanner_with_archive(t *testing.T) {
 		require.NotNil(t, results)
 
 		failed := results.GetFailed()
-		assert.Equal(t, 26, len(failed))
+		assert.Equal(t, 25, len(failed))
```
owenrumney, line 50: how come we lost one?

owenrumney (approved):

pkg/scanners/helm/parser/parser.go
liamg: :broom: :broom:

Yeah, test failed as i fixed the rego getting run twice so the number was different, so noticed the output

pkg/scanners/helm/test/scanner_test.go
liamg: i fixed the rego getting run twice so the number was different

liamg: i had a bug where if you had 2 rules called `deny` it would run them both twice

PR #589, opened 2022-05-10
fix: detection of first/last cause line

liamg:


owenrumney (approved):
Checks watch :clock1:

liamg:
borked it

PR #592, opened 2022-05-11
chore: peek inside tar to see if its helm

owenrumney:


pkg/detection/detect.go
```diff
@@ -61,17 +86,19 @@ func init() {
...
 			contents := make(map[string]interface{})
-			if err := json.Unmarshal(data, &contents); err == nil {
+			err = json.Unmarshal(data, &contents)
+			if err == nil {
 				if _, ok := contents["terraform_version"]; ok {
 					_, stillOk := contents["format_version"]
 					return stillOk
 				}
 			}
+			_ = err
```
liamg, line 96:
```suggestion
```

pkg/scanners/kubernetes/parser/parser.go
```diff
@@ -91,7 +91,10 @@ func (p *Parser) required(fs fs.FS, path string) bool {
 		return false
 	}
 	defer func() { _ = f.Close() }()
-	return detection.IsType(path, f, detection.FileTypeKubernetes)
+	if data, err := ioutil.ReadAll(f); err == nil {
```
liamg, line 94: I think we can pass `f` to the `IsType` function without this middle reader?

owenrumney: f doesn't implement`io.ReaderSeeker` which is expected by `IsType`

liamg (approved):

PR #603, opened 2022-05-16
Always omit Highlighted

knqyf263:


owenrumney (changes requested):
That doesn't work. I've already tried it.

Something inside trivy or fanal is serializing/deserialzing and the value gets lost in the process

PR #611, opened 2022-05-18
fix: Fix false positives for KSV038

### LINKED ISSUE #606: KSV038 is triggered and failing on all Kubernetes files expect of good network security policies
Used by trivy:0.28
No negative test in selector_usage_in_network_policies_test.rego

Alexander-Bartosh:
Closes #606


liamg (dismissed):
Awesome, thank you!

PR #614, opened 2022-05-20
feat: handle multiple charts at the same time

owenrumney:


pkg/scanners/helm/scanner.go
```diff
@@ -89,7 +90,49 @@ func (s *Scanner) SetPolicyFilesystem(policyFS fs.FS) {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if path == "." {
```
liamg, line 105:
```suggestion
		if d.IsDir() {
```
Am I right in thinking we can skip all directories here? As we only want archives or `Chart.yaml`s?

liamg (changes requested):

pkg/scanners/helm/scanner.go
owenrumney: done

liamg (approved):

PR #628, opened 2022-05-27
fix: support source/target tags for GCP firewall

### LINKED ISSUE #577: bug: GCP firewall rule between tags get flagged as google-compute-no-public-ingress
**Describe the bug**
False positive `google-compute-no-public-ingress` on firewall rule between two GCP network tags.

**To Reproduce**
1. Create a GCP ingress firewall rule between two network tags (in my multiple VMs to one VM).
2. Run official tfsec docker image
3. Get `google-compute-no-public-ingress` error

**Expected behavior**
My understanding is, that network tags are connected to VMs [GCP network tag docs](https://cloud.google.com/vpc/docs/add-remove-network-tags) (I might be wrong about this assumption though).
So a firewall rule between two tags can never be about public ingress.

That is why I consider this a false positive.

**Screenshots/Output**
Output of  tfsec v1.19.1:
Screenshot from 2022-05-09 09-09-28

**System Info**
 - official docker image aquasec/tfsec:v1.19.1
 - Terraform version: 1.1.X
 - GitLab-CI

**Example Code**
Redacted example
```
resource "google_compute_firewall" "allow-vms-to-some-machine" {
  name      = "allow-vms-to-some-machine"
  network   = local.network
  priority  = 1300
  direction = "INGRESS"
  allow {
    protocol = "tcp"
    ports    = ["8081"]
  }
  source_tags = ["vms"]
  target_tags = ["some-machine"]
}
```

**Additional context**
-

owenrumney:
Resolves #577

### DIFF
--- a/pkg/terraform/attribute.go
+++ b/pkg/terraform/attribute.go
@@ -114,6 +114,18 @@ func (a *Attribute) AsStringValueOrDefault(defaultValue string, parent *Block) t
 	)
 }
 
+func (a *Attribute) AsStringValueSliceOrEmpty(parent *Block) (stringValues []types.StringValue) {
+	if a.IsNil() {
+		return stringValues
+	}
+
+	for _, v := range a.ValueAsStrings() {
+		stringValues = append(stringValues, types.String(v, parent.GetMetadata()))
+	}
+
+	return stringValues
+}
+
 func (a *Attribute) AsBoolValueOrDefault(defaultValue bool, parent *Block) types.BoolValue {
 	if a.IsNil() {
 		return types.BoolDefault(defaultValue, parent.GetMetadata())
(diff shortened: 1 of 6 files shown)


pkg/terraform/attribute.go
```diff
@@ -114,6 +114,18 @@ func (a *Attribute) AsStringValueOrDefault(defaultValue string, parent *Block) t
 	)
 }
 
+func (a *Attribute) AsStringValueSliceOrEmpty(parent *Block) (stringValues []types.StringValue) {
+	if a.IsNil() {
+		return stringValues
+	}
+
+	for _, v := range a.ValueAsStrings() {
+		stringValues = append(stringValues, types.String(v, parent.GetMetadata()))
```
liamg, line 123: I think it should be safe to get the metadata from the attribute itself if it's set:
```suggestion
		stringValues = append(stringValues, types.String(v, a.GetMetadata()))
```

liamg (changes requested):

liamg (approved):

Issue #669, opened 2022-06-07
bug: PSS rego issues


joebowbeer:
List of issues with [rule/kubernetes/policies/pss](https://github.com/aquasecurity/defsec/tree/master/internal/rules/kubernetes/policies/pss):

baseline vs https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline

* [x] Missing: rule for `windowsOptions.hostProcess` https://github.com/aquasecurity/defsec/pull/686
* [x] Missing: baseline rule for seccomp profile: seccompProfile.type must not be explicitly set to Unconfined (but may be undefined) https://github.com/aquasecurity/defsec/pull/686

restricted vs https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

* [x] Bug: #668 volume type csi _should_ be allowed
* [x] Missing: check runAsUser != 0, though there is a _general_ rule that checks runAsUser > 10000 https://github.com/aquasecurity/defsec/pull/688
* [x] Extraneous: pss/restricted contains a rule for runAsGroup that is not specified in doc https://github.com/aquasecurity/defsec/pull/688
* [x] Missing: check for explicit seccompProfile type. The doc says that the absence of a seccomp profile type is prohibited, but the rego rule does not deny it https://github.com/aquasecurity/defsec/pull/688
* [x] Missing: check that containers drop ALL capabilities and only add back the NET_BIND_SERVICE capability if needed; there are two rules outside of pss that check for dropped capabilities, but there is no rule verifying that NET_BIND_SERVICE is the only capability added back https://github.com/aquasecurity/defsec/pull/688

liamg:
Thanks @joebowbeer - I'm looking at sorting this now.

PR #686, opened 2022-06-08
fix(kubernetes): Add missing PSS baseline policies

liamg:
Relates to #669


internal/rules/kubernetes/policies/pss/baseline/10_windows_host_process.rego
```diff
@@ -0,0 +1,38 @@
+package builtin.kubernetes.KSV103
```
owenrumney, line 1: Do you know why we jump for 40 - 102? 

Chen is using 41+ in his PR. I'm not clear where 102 came from

liamg: Nope :s 

I'm using `make id` to pick my IDs now.

PR #692, opened 2022-06-13
fix: Issues with cache directory

liamg:
This fixes a few issues with the cache directory:

- The cache dir is no longer created until/unless it is needed (e.g. no longer created when `tfsec --help` is run etc.)
- Cache directory is now named `.aqua` to avoid confusion when `.tfsec` was being created in Trivy.
- Project directory is no longer tainted by the cache dir, causing issues in CI/docker etc.


owenrumney:
This resolves aquasecurity/tfsec#1705, right?

owenrumney (approved):

liamg:
Yep :+1: I'll link it to the PR where we update defsec in tfsec :+1:

PR #704, opened 2022-06-21
fix: Handle helm charts with templated names

### LINKED ISSUE #703: yaml parsing error
Hi! we have some issue )

the customer gets an error (https://github.com/aquasecurity/trivy/issues/2362) when try to scan a helm chart with variables:
the error:
```sh
yaml: invalid map key: map[string]interface {}{"COMPONENT_NAME":interface {}(nil)}
```
`Chart.yaml`:
```yaml
apiVersion: v2
name: {{COMPONENT_NAME}}
version: {{COMPONENT_VERSION}}
description: A Helm chart for Kubernetes
keywords:
- kublr
- audit
```
The problem appears here:
https://github.com/aquasecurity/defsec/blob/0c2778fa5a4f6afb598f25807d597cc16f7af0ac/pkg/scanners/helm/parser/parser.go#L136-L139

Could you look into it? thanks a lot!

liamg:
Resolves #703

### DIFF
--- a/pkg/scanners/helm/parser/parser.go
+++ b/pkg/scanners/helm/parser/parser.go
@@ -13,6 +13,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/google/uuid"
+
 	"github.com/aquasecurity/defsec/pkg/debug"
 
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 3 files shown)


owenrumney:
So are we just stepping over it if it doesn't have a resolvable name?

liamg:
Yeah, and using the name of the directory as the name, or failing that a random uuid.

owenrumney (approved):

afdesk:
Hi! I also tried to resolve this issue and fount next question: https://stackoverflow.com/questions/63270815/is-there-a-way-to-validate-parse-a-yaml-file-with-go-template-variables-in-it

just FYI.

PR #706, opened 2022-06-23
feat: Add json and markdown format to --run-statistics

amandahla:
Proposal: add formats json and markdown to statistics output

Why: https://github.com/aquasecurity/tfsec/issues/1790

### DIFF
--- a/pkg/scanners/terraform/executor/statistics.go
+++ b/pkg/scanners/terraform/executor/statistics.go
@@ -1,6 +1,8 @@
 package executor
 
 import (
+	"encoding/json"
+	"fmt"
 	"io"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


pkg/scanners/terraform/executor/statistics.go
```diff
@@ -12,27 +14,54 @@ import (
...
 	sorted := SortStatistics(statistics)
+
+	if format == "json" {
+		result := StatisticsResult{Result: sorted}
+		val, err := json.MarshalIndent(result, "", "    ")
+		if err != nil {
+			return err
+		}
+
+		fmt.Println(string(val))
```
liamg, line 51: We have a writer to send the output to here, i.e.
```suggestion
		_, _ = fmt.Fprintln(w, string(val))
```

liamg (changes requested):
Thanks @amandahla this is great - I've added one minor comment then we're good to merge :+1:

liamg (approved):

liamg:
I've updated the branch with the minor change, hope this is ok :)

amandahla:
> I've updated the branch with the minor change, hope this is ok :)

Thanks @liamg , sorry for the delay!

PR #725, opened 2022-07-07
Update liamg/iamgo to 0.0.9

sshearn:
There's an issue when policies with multiple statements without SIDs don't get included. It esstentially will only include the last statement without an SID. Terraform / AWS allows this blank SIDs so change needs to be included in order to correctly parse all statements. 

This is fixed in liamg/iamgo v0.0.9


liamg:
Thank you @sshearn :+1:

liamg (approved):

PR #734, opened 2022-07-19
feat: Add framework classification capability for rules

liamg:
Backward compatible.

All rules are registered to the `default` framework by default. Specifying anything in the list of frameworks for a rule will cause it to be removed from the `default` framework unless `default` is explicitly specified in the list.

Rules can be requested for zero or more frameworks. Rules are returned which exist in any of the the supplied frameworks. If no frameworks are requested, rules for the `default` framework will be returned.

### DIFF
--- a/internal/rules/register.go
+++ b/internal/rules/register.go
@@ -90,7 +90,7 @@ func (r *registry) deregister(rule RegisteredRule) {
 	}
 }
 
-func (r *registry) GetFrameworkRules(fw ...framework.Framework) []RegisteredRule {
+func (r *registry) getFrameworkRules(fw ...framework.Framework) []RegisteredRule {
 	r.RLock()
 	defer r.RUnlock()
 	var registered []RegisteredRule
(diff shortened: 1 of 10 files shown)


internal/rules/register.go
```diff
@@ -30,36 +29,93 @@ func (r RegisteredRule) Evaluate(s *state.State) scan.Results {
...
+func (r *registry) Reset() {
+	r.Lock()
+	defer r.Unlock()
+	r.frameworks = make(map[framework.Framework][]RegisteredRule)
 }
 
-func GetRegistered() []RegisteredRule {
-	return registeredRules
+func GetFrameworkRules(fw ...framework.Framework) []RegisteredRule {
+	return coreRegistry.GetFrameworkRules(fw...)
```
owenrumney, line 120: do you think if `fw` is empty, `fw = []framwork.Framwork{framework.Default}`?

owenrumney: oh ignore, just saw it further up

owenrumney (dismissed):
That was quick :rocket:

owenrumney (approved):

PR #762, opened 2022-07-25
fix: add configmap support to k8s scanner

josedonizetti:
Signed-off-by: Jose Donizetti <jdbjunior@gmail.com>

Add support to kubernetes configmap detection. Instead of testing specifically for ConfigMap, I opted to only test for "apiVersion", "kind", "metadata" to detect if a yaml is k8s or not, though I'm it is ignoring rbac because it has its own scanner. This means, others resources would also be detected as K8S, like secrets, endpoints, endpointslices etc, which I think it is okay because they are k8s objects. So, actually what this PR does is to stop detecting only workloads. 

Let me know what you think please. We want trivy-operator to be able to report on ConfigMaps.

### DIFF
--- a/pkg/detection/detect.go
+++ b/pkg/detection/detect.go
@@ -174,13 +174,19 @@ func init() {
 			return false
 		}
 
-		expectedProperties := []string{"apiVersion", "kind", "metadata", "spec"}
+		expectedProperties := []string{"apiVersion", "kind", "metadata"}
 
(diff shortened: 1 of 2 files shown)


chen-keinan (approved):

chen-keinan:
LGTM! just one comment , I would ad another test with `Role` yaml to make sure it doesn't catch it

owenrumney (approved):

PR #764, opened 2022-07-26
feat: Add framework references specification option for all rules

liamg:


owenrumney (approved):
:building_construction:

PR #767, opened 2022-07-27
feat: Add exclude-ignores flag and --config-file attribute

alexandrupopafc:
This is intended for adding `exclude-ignores` flag and --config-file attribute.

In some cases, we want to enforce some rules and remove the ability to `#tfsec:ignore`

Follow up PR for tfsec:
https://github.com/aquasecurity/tfsec/pull/1839


liamg (changes requested):
Thanks @alexandrupopafc - looks good. It seems you need to run `goimports` on `pkg/scanners/terraform/executor/executor.go`, then we're good to merge :+1:

liamg (approved):

PR #780, opened 2022-08-01
fix(rego): Allow quotes in working directories for Dockerfiles

liamg:
Relates to https://github.com/aquasecurity/trivy/issues/2622


owenrumney (approved):
LGTM

PR #832, opened 2022-08-10
feat: add concurrency for cloud scanning

owenrumney:
- feat: parrallelise the S3 bucket get
- feat: parralel processing for s3 and iam
- feat: make users parrallel too


liamg (approved):
:dash:

PR #837, opened 2022-08-10
fix: clean up the debug logging for s3

owenrumney:
debug logging is pretty vocal on the errors, clean up by checking the
code isn't the standard response


liamg (approved):
:+1:

Issue #719, opened 2022-07-04
Map ignored results as "skipped" in JUnit


askingcat:
Currently any [ignored](https://github.com/aquasecurity/defsec/blob/872e1c2668813a9a0730c80ae6266fe2ed37db73/pkg/scan/result.go#L19) scan result will be mapped as a [failure](https://github.com/aquasecurity/defsec/blob/872e1c2668813a9a0730c80ae6266fe2ed37db73/pkg/formatters/junit.go#L96-L121) for JUnit. 

This might be a bit confusing and can make it more difficult to integrate other tools. 

While it would not be a perfect fit, I wonder whether it would make more sense to map "ignored" as "skipped" for JUnit. If you are OK with such a change, I would be happy to try to create a PR for this :-).

owenrumney:
Hey @askingcat  - still willing to PR this? I'm happy with the proposed solution :+1:

askingcat:
Hey @owenrumney,

sure :smile:.

askingcat:
The merged code is already included in [tfsec 1.27.2](https://github.com/aquasecurity/tfsec/releases/tag/v1.27.2) and I just confirmed that it works :slightly_smiling_face:.

PR #879, opened 2022-08-18
feat: distinguish failed and ignored for JUnit (#719)

askingcat:
Previously all not passed scan results were treated as a failure by the
JUnit formatter. This was a problem for e.g. tfsec, since with a JUnit
result viewer there was no way to distinguish between ignored tests and
actually failed ones.

Now "ignored" scan results are marked as "skipped".

Additionally, a test for passed scans has been added as well. It already
worked as intended, but so far we had no test to ensure this.


liamg (approved):
LGTM, thank you!

Issue #925, opened 2022-09-06
bug:  Load the k8s Embedded Libraries by default


chen-keinan:
it is required to load the `k8s Embedded Libraries` by default without loading the `embedded policies`

**The motivation:**
`Trivy-operator` offer the ability to run custom policies.  we recommend users to follow our conversion for writing policies meaning to use the k8s Embedded Libraries when they write the custom policies.

in some cases the users  do not want to use our embedded policies , so if they run custom policies without the embedded policies the custom policy `eval` fails due to missing embedded lib.
it will be great to support it.

Let me know if this req. doesn't make sense 😵‍💫

PR #926, opened 2022-09-06
feat: load rego embedded libs by default

chen-keinan:
Close: #925


owenrumney (approved):

simar7:
hi @chen-keinan - we're trying to bring back policy fetching into trivy for which we'll need to disable embedded policies and libraries, if a new bundle is found upstream. Context is here: https://github.com/aquasecurity/defsec/commit/d128e5289f6ec2e3120216a9953f7f34ecea9658#r86794387

tldr: Currently this behaviour doesn't allow trivy to download a bundle (with both policies and libraries in it) as libraries are unconditionally embedded into the trivy binary, causing rego namespace collisions. Bundle will include both policies and libraries. 

I understand you had made this change to solve https://github.com/aquasecurity/defsec/issues/925 - is it possible we (for the users of trivy operator) can provide the libraries in another way? 

Ideally, it'd be best to keep policies and libraries together behind the same flag. In other words, `if loadEmbedded {` should gate both the policies and libraries as decoupling the two introduces situations where you can have potentially have a mismatch (old libraries, new policies) or rego namespace collisions (this scenario). 

Thoughts?

simar7:
cc @liamg @itaysk

simar7:
Created https://github.com/aquasecurity/defsec/pull/1012 to address this.

chen-keinan:
> hi @chen-keinan - we're trying to bring back policy fetching into trivy if a new bundle is found upstream. This is so that we can decouple trivy releases from policy updates.
> 
> tldr: Currently this behaviour doesn't allow trivy to download a bundle (with both policies and libraries in it) as libraries are unconditionally embedded into the trivy binary, causing rego namespace collisions. Bundle will include both policies and libraries. Context is here: [d128e52#r86794387](https://github.com/aquasecurity/defsec/commit/d128e5289f6ec2e3120216a9953f7f34ecea9658#r86794387)
> 
> I understand you had made this change to solve #925 - is it possible we (for the users of trivy operator) can provide the libraries in another way? Maybe they could be stored as configmaps within the operator deployment charts? Just a thought.
> 
> Ideally, it'd be best to keep policies and libraries together behind the same flag. In other words, `if loadEmbedded {` should gate both the policies and libraries as decoupling the two introduces situations where you can have potentially have a mismatch (old libraries, new policies) or rego namespace collisions (this scenario).
> 
> Thoughts?

sure , lets discuss the options

simar7:
> > hi @chen-keinan - we're trying to bring back policy fetching into trivy if a new bundle is found upstream. This is so that we can decouple trivy releases from policy updates.
> > tldr: Currently this behaviour doesn't allow trivy to download a bundle (with both policies and libraries in it) as libraries are unconditionally embedded into the trivy binary, causing rego namespace collisions. Bundle will include both policies and libraries. Context is here: [d128e52#r86794387](https://github.com/aquasecurity/defsec/commit/d128e5289f6ec2e3120216a9953f7f34ecea9658#r86794387)
> > I understand you had made this change to solve #925 - is it possible we (for the users of trivy operator) can provide the libraries in another way? Maybe they could be stored as configmaps within the operator deployment charts? Just a thought.
> > Ideally, it'd be best to keep policies and libraries together behind the same flag. In other words, `if loadEmbedded {` should gate both the policies and libraries as decoupling the two introduces situations where you can have potentially have a mismatch (old libraries, new policies) or rego namespace collisions (this scenario).
> > Thoughts?
> 
> sure , lets discuss the options

Previously, in the case of Trivy for using custom policies the users have to supply their own libraries in case they need them. Having said that, maybe we can instead create a reference deployment (example helm chart, for instance) that includes some scaffolding to help them?

Long term, decoupling policies from Trivy binary helps us as policies get updated often compared to Trivy itself. This is the same model that Trivy DB follows for updating vulnerability definitions. 

What do you think? Maybe we can jump on a call and sort it out.

chen-keinan:
> 
> What do you think? Maybe we can jump on a call and sort it out.

sure I'll be happy to have a call to discuss it

PR #953, opened 2022-09-21
feat: add OnlyContains helper function

owenrumney:

### DIFF
--- a/pkg/terraform/attribute.go
+++ b/pkg/terraform/attribute.go
@@ -8,9 +8,8 @@ import (
 	"strconv"
 	"strings"
 
-	defsecTypes "github.com/aquasecurity/defsec/pkg/types"
-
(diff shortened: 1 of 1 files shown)


pkg/terraform/attribute.go
```diff
@@ -483,6 +482,60 @@ func (a *Attribute) Contains(checkValue interface{}, equalityOptions ...Equality
 	return strings.Contains(val.AsString(), stringToLookFor)
 }
 
+func (a *Attribute) OnlyContains(checkValue interface{}) bool {
```
liamg, line 485: We could do `checkValue ...interface{}` so you can do `OnlyContains(x)` or `OnlyContains(x, y, z)` - WDYT?

liamg: (then we can remove the check for `checkValue.([]interface{})` on line 494 too

liamg (approved):

PR #960, opened 2022-09-27
feat: Rego engine QoL improvements

### LINKED ISSUE #951: feat: Core rego engine rework and QoL improvements
- [x] Rule metadata (`__rego_metadata__`) should be parsed from the [official rego annotation format](https://www.openpolicyagent.org/docs/latest/annotations/).
- [x] We should also use annotations to replace `__rego_input__` to simplify and standardise the rule code.
- [x] Allow full customisation via new metadata - as in https://github.com/aquasecurity/defsec/issues/955
- [x] The rego input schema should be documented with json schema and applied in annotations for all scanners (e.g. dockerfile, terraform, aws etc.)
  - [x] Dockerfile
  - [x] Kubernetes
  - [x] RBAC
  - [x] Cloud State (blocked by https://github.com/open-policy-agent/opa/issues/5166)
- [x] Helper methods should be made available to ensure the rule authoring process is as simple as possible, like the Go rule experience.
- [x] The defsec-specific `data.lib` code should be replaced with Go functions that are made available to the rego at runtime.
- [ ] Update rego metadata docs
- [ ] Add schema to docs, and generate documentation from all json schema documents
- [x] ~PR to OPA to add remote schema support - see https://github.com/open-policy-agent/opa/issues/4478~
- [x] ~Link remote schema in check metadata~
- [x] Fix https://github.com/aquasecurity/tfsec/issues/1901
- [x] Fix https://github.com/open-policy-agent/opa/issues/5166

This is a prerequisite for https://github.com/aquasecurity/trivy/issues/2917

liamg:
Resolves #951


internal/rules/defsec/lib/defsec.rego
```diff
@@ -1,15 +0,0 @@
...
-		"msg": msg,
-		"startline": object.get(metadata, "startline", object.get(metadata, "StartLine", 0)),
-		"endline": object.get(metadata, "endline", object.get(metadata, "EndLine", 0)),
-		"filepath": object.get(metadata, "filepath", object.get(metadata, "Path", "")),
-		"explicit": object.get(metadata, "explicit", false),
-		"managed": object.get(metadata, "managed", true),
-		"fskey": object.get(metadata, "fskey", ""),
-		"resource": object.get(metadata, "resource", ""),
-	}
-}
```
simar7, line 15: sorry just stumbled my way here – will this helper live somewhere else? (Maybe it already does and I might have missed it) 

Today there are references to this in other policies such as https://github.com/aquasecurity/defsec/blob/master/internal/rules/docker/policies/add_instead_of_copy.rego#L36

We probably also need to update the policies with the correct import of wherever this helper would end up living as today they don't seem to have them.

I ran into this while importing all defsec rules into trivy through the OCI artifact. 
```
docker/policies/add_instead_of_copy.rego:36: rego_type_error: undefined function result.new
```

liamg: So this now works without the import as the function is provided as an extension via the Go library - it is now defined in Go.

PR #966, opened 2022-09-29
fix: Support absolute paths for policies in filesystems

liamg:


pkg/rego/load.go
```diff
@@ -16,10 +16,14 @@ func isRegoFile(name string) bool {
 	return strings.HasSuffix(name, bundle.RegoExt) && !strings.HasSuffix(name, "_test"+bundle.RegoExt)
 }
 
+func sanitisePath(path string) string {
```
owenrumney, line 19: I wonder if this would have issues with a samba share, eg; `//share/rego-files`

liamg: This should only apply inside an `fs.FS` - though there could well be issues with scanning sambsa shares further up in the code I guess :vomiting_face:

owenrumney (approved):

PR #968, opened 2022-09-30
feat: Add arm scanner

owenrumney:

### DIFF
--- a/pkg/types/metadata.go
+++ b/pkg/types/metadata.go
@@ -14,6 +14,7 @@ type Metadata struct {
 	isExplicit     bool
 	isUnresolvable bool
 	parent         *Metadata
+	internal       interface{}
 }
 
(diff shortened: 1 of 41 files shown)


pkg/types/metadata.go
```diff
@@ -124,6 +129,15 @@ func (m Metadata) Root() Metadata {
 	return *meta
 }
 
+func (m Metadata) WithInternal(internal interface{}) Metadata {
+	m.internal = internal
```
liamg, line 133: Can we get rid of these `Internal` functions now? I can't remember what they were for but I don't think we're using them now?

liamg (changes requested):
Looks good, hoping we can delete the `SetInternal` thing though :crossed_fingers:

liamg (approved):
ignore last comment

PR #987, opened 2022-10-05
fix: Add missing reference URLs to all rego policies

liamg:


pkg/rego/metadata.go
```diff
@@ -213,6 +213,19 @@ func (m *MetadataRetriever) updateMetadata(meta map[string]interface{}, metadata
 			metadata.Frameworks[framework.Framework(fw)] = sections
 		}
 	}
+	if raw, ok := meta["related_resources"]; ok {
```
owenrumney, line 216: I have a feeling this change will increase the cyclomatic compexity :thinking:

liamg: oh no

owenrumney (approved):

PR #996, opened 2022-10-07
fix: return empty array with no warnings in --format json

koka831:
While developing a vim plugin, I noticed that when I run `tfsec` with the `--format json` option, if there is no lint error, it returns the following response.

```json
{ "results": null }
```

Although `null` is valid value in the JSON schema, it seems more appropriate to return an empty array so that we can handle the output more easily.

## changes

* use empty array instead of null when there's no warnings
* add test to check JSON formatter handles empty results


owenrumney (approved):
Looks good thanks. I'd seen the tweet and you've saved me looking into it further.#

Thanks for the contribution.

PR #1021, opened 2022-10-18
feat: add compliance spec for nsa

chen-keinan:
## Description:
Add compliance spec for nsa

Close: #1022


liamg (approved):

owenrumney (approved):

internal/specs/compliance/nsa-1.0.yaml
```diff
@@ -0,0 +1,175 @@
+---
```
simar7, line 1: For my understanding, how was this spec created? Is it available somewhere as a yaml file or you just translated their PDF into this yaml?

Would you know if they have similar spec for AWS?

chen-keinan: I have created it my self bases on investigation of the NSA spec and made the relevant mapping to rego

internal/specs/compliance/nsa-1.0.yaml
```diff
@@ -0,0 +1,175 @@
...
+      description: 'Control check whether RBAC permission is in use'
+      id: '7.1'
+      checks:
+        - id: AVD-KCV-0008
+      severity: 'CRITICAL'
+    - name: Audit policy is configure
+      description: 'Control check whether audit policy is configure'
+      id: '8.0'
+      defaultStatus: 'FAIL'
+      checks:
```
simar7, line 162: Another quick q: why some of these checks don't have an associated rule (check)? Is it because we don't support it ?

chen-keinan: Yes , see #1042 ,we now tag it as Manual checks. Trivy can't run these checks in a non intrusive mode (outside of the cluster) .but it still part of Nsa so we added it.

PR #1025, opened 2022-10-22
feat(aws): add msk encryption at rest check

simar7:
Resolves: #458


liamg (dismissed):
Looks good, though we should probably also add support for this in AWS live account scanning, see https://github.com/aquasecurity/defsec/blob/master/internal/adapters/cloud/aws/msk/adapt.go

owenrumney (approved):

PR #1045, opened 2022-11-04
feat(compliance): Support checks by spec

simar7:
Addresses: https://github.com/aquasecurity/trivy/issues/2919


owenrumney (dismissed):

internal/rules/register.go
```diff
@@ -113,6 +116,31 @@ func (r *registry) getFrameworkRules(fw ...framework.Framework) []RegisteredRule
...
 }
 
+func (r *registry) getSpecRules(spec string) []RegisteredRule {
+	r.RLock()
+	defer r.RUnlock()
+	var specRules []RegisteredRule
+
+	var complianceSpec cs.ComplianceSpec
+	specContent := specs.GetSpec(spec)
+	_ = yaml.Unmarshal([]byte(specContent), &complianceSpec)
```
chen-keinan, line 126: error here is ignored ?

pkg/scanners/cloud/aws/scanner.go
```diff
@@ -186,6 +191,16 @@ func (s *Scanner) Scan(ctx context.Context, cloudState *state.State) (results sc
...
 }
 
+func (s *Scanner) getRegisteredRules() []rules.RegisteredRule {
+	var regRules []rules.RegisteredRule
+	if len(s.frameworks) > 0 { // Only for maintaining backwards compat
+		regRules = rules.GetFrameworkRules(s.frameworks...)
+	} else {
+		regRules = rules.GetSpecRules(s.spec)
+	}
+	return regRules
```
chen-keinan, line 201: nit :
```suggestion
	if len(s.frameworks) > 0 { // Only for maintaining backwards compat
		return regRules = rules.GetFrameworkRules(s.frameworks...)
	} 
	return regRules = rules.GetSpecRules(s.spec)

```

chen-keinan:
@simar7 looks good 🚀  I have added couple of comments

internal/rules/register.go
simar7: that's right - I ignored it for a couple of reasons:

1. We're unmarshaling something we've embedded ourselves. (no user input)
2. I wasn't sure what we should do if we encounter an error anyway. We could early return an empty slice back. 

What do you think?

pkg/scanners/cloud/aws/scanner.go
simar7: thanks, fixed https://github.com/aquasecurity/defsec/pull/1045/commits/98aa205753a4005f99f5e001aad3741db26e19a6

internal/rules/register.go
chen-keinan: I think we should handle error in some way even if we put the json by our self as we also can make a mistake.
anyone might want to use this method in the future should know its safe to use.

in case of not handle error we could get nil pointer at least we should protect from nil pointer :
```for _, csRule := range complianceSpec.Spec.Controls {```

simar7: Thanks good point, I added it here https://github.com/aquasecurity/defsec/pull/1045/commits/be774ff1f1025cb560ceffd3bc8c1264e473d418

Also another test should be able to cover the case of bogus spec from a loading point of view https://github.com/aquasecurity/defsec/pull/1045/files#diff-31f3b00f062e6251cb5439c6086f5e5a745d7caefda6db781101450933a2b354R19

chen-keinan (approved):

PR #1056, opened 2022-11-16
Added policy to check that all region is enable or not in cloudtrail

aqibbutt20:


simar7:
hi @aqibbutt20 thanks for the PR - could you please sign the CLA for us?

simar7:
> @simar7 defsec [already checks](https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/cloudtrail/enable_all_regions.go) this rule.

Yes the plan is to re-write existing Go rules in Rego. 

As part of this PR, we should also deprecate the Go rules. Otherwise we will have a finding twice.

PR #1058, opened 2022-11-16
Compatible with Go 1.18

knqyf263:
Aqua internal tools need to compile Trivy/defsec with Go 1.18. We need to backport `net/url.JoinPath`, which is introduced in Go 1.19 so defsec will be compatible with Go 1.18.
https://pkg.go.dev/net/url#JoinPath

### DIFF
--- a/pkg/scanners/azure/functions/uri.go
+++ b/pkg/scanners/azure/functions/uri.go
@@ -1,15 +1,29 @@
 package functions
 
-import "net/url"
+import (
+	"net/url"
+	"path"
+)
 
 func Uri(args ...interface{}) interface{} {
 	if len(args) != 2 {
 		return ""
 	}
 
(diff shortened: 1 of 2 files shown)


simar7 (commented):
It lgtm on the first glance, I added another test case as well. But there are a variety of test cases that url.JoinPath() tests. 

BTW Do we also need change the go.mod for defsec? https://github.com/aquasecurity/defsec/blob/master/go.mod#L3

knqyf263:
I didn't add support for those cases as they are edge cases and this is a temporal workaround. But it is surely better to support them. Thanks!

It should work without changing go.mod. Might be safer to change it though.

simar7 (approved):
You're right, I confused the go.mod version with something else. I'm fine to keep the go.mod as is for now.

PR #1059, opened 2022-11-16
Added policy to check log_group_customer_key for cloudwatch

SanaaYousaf:


simar7:
Does this rule currently exist? https://github.com/aquasecurity/defsec/blob/master/internal/rules/aws/cloudwatch/log_group_customer_key.go

SanaaYousaf:
> Does this rule currently exist? https://github.com/aquasecurity/defsec/blob/master/internal/rules/aws/cloudwatch/log_group_customer_key.go

Yes but now we are adding rego for that rule as we also need to convert go rules to rego

simar7:
> > Does this rule currently exist? https://github.com/aquasecurity/defsec/blob/master/internal/rules/aws/cloudwatch/log_group_customer_key.go
> 
> Yes but now we are adding rego for that rule as we also need to convert go rules to rego

I believe as we move over rules from Go to Rego, we should remove them in Go as otherwise we'll have 2 different rules flagging the same problem causing duplicate results. 

WDYT @itaysk?

itaysk:
Agree, about replacing go rules with rego. I'm not sure about reusing IDs though.

simar7:
hi @SanaaYousaf - I just merged https://github.com/aquasecurity/defsec/pull/1088, this will let you add existing documentation to rego policy metadata from the existing Go rules. In particular the Terraform and Cloud Formation help texts that are currently here: https://github.com/aquasecurity/defsec/blob/master/internal/rules/aws/cloudwatch/log_group_customer_key.go#L24-L35


For an example of how to add and structure this metadata please see my comment here: https://github.com/aquasecurity/defsec/pull/1088#issue-1467304051 

We should do this for the rules that are being ported from Go to Rego. In addition once the rego metadata is populated, we can delete the associated *cf.go, *tf.go files for the Go rule.

For now, let's keep the AVD IDs to be unique (what you currently have) instead of reusing them. Ping me if you need any other help.

simar7:
hi @SanaaYousaf any update on this?

PR #1060, opened 2022-11-17
Enable option to create custom scanner

tamirkiviti13:
Today there is no option to create a custom scanner for a new type of source.
"trivy-plugin-aqua" need this option, so this PR enables the option to do so.


pkg/rego/embed.go
```diff
@@ -29,6 +29,11 @@ func init() {
 		modules[name] = policy
 	}
 
+	RegisterRegoRules(modules)
+}
+
+// This function is exported for trivy-plugin-aqua purposes only
+func RegisterRegoRules(modules map[string]*ast.Module) {
```
simar7, line 36:
```suggestion
func RegisterRegoRules(modules map[string]*ast.Module) {
```

pkg/rego/embed.go
```diff
@@ -61,14 +66,15 @@ func init() {
...
 }
 
 func loadEmbeddedLibraries() (map[string]*ast.Module, error) {
-	return recurseEmbeddedModules(rules.EmbeddedLibraryFileSystem, ".")
+	return RecurseEmbeddedModules(rules.EmbeddedLibraryFileSystem, ".")
 }
 
-func recurseEmbeddedModules(fs embed.FS, dir string) (map[string]*ast.Module, error) {
+// This function is exported for trivy-plugin-aqua purposes only
+func RecurseEmbeddedModules(fs embed.FS, dir string) (map[string]*ast.Module, error) {
```
simar7, line 77:
```suggestion
func RecurseEmbeddedModules(fs embed.FS, dir string) (map[string]*ast.Module, error) {
```

simar7 (approved):
lgtm

Issue #1067, opened 2022-11-18
bug: S3 bucket encryption check is incorrect.


int-tt:
**Describe the bug**

There is a mistake in the logic of S3 bucket encryption, which can be detected incorrectly even when bucket encryption is enabled.
**To Reproduce**
Using server-side-encryption-configuration. 
https://avd.aquasec.com/misconfig/aws/s3/avd-aws-0088/
```
{
  "Rules": [
    {
      "ApplyServerSideEncryptionByDefault": {
        "SSEAlgorithm": "AES256"
      }
    }
  ]
}
```
**Expected behavior**

This setting should be passable.

The `BucketKeyEnabled` used to determine whether bucket encryption is enabled is a confirmation of the bucket key's validity.
https://github.com/aquasecurity/defsec/blob/master/internal/adapters/cloud/aws/s3/s3.go#L196
https://docs.aws.amazon.com/ja_jp/AmazonS3/latest/API/API_ServerSideEncryptionRule.html

Even if it is not enabled, if it is specified in `ServerSideEncryptionByDefault` and `SSEAlgorithm` , the bucket is encrypted.

**Output of your tfsec command with --debug flag**

**System Info**

 - trivy: 0.34.0
 - tfsec version: `?`
 - terraform version: `?`
 - OS: `?`
**Example Code**

**Additional context**

PR #1068, opened 2022-11-18
fix: bucket encryption is enabled if SSEAlgorithm is specified

int-tt:
Resolves: https://github.com/aquasecurity/defsec/issues/1067

### DIFF
--- a/internal/adapters/cloud/aws/s3/s3_test.go
+++ b/internal/adapters/cloud/aws/s3/s3_test.go
@@ -338,6 +338,7 @@ func bootstrapBucketEncryption(t *testing.T, api *s3api.Client, ctx context.Cont
 					ApplyServerSideEncryptionByDefault: &s3types.ServerSideEncryptionByDefault{
 						SSEAlgorithm: s3types.ServerSideEncryptionAes256,
 					},
+					BucketKeyEnabled: true,
 				},
 			},
 		},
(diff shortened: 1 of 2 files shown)


internal/adapters/cloud/aws/s3/s3_test.go
```diff
@@ -338,7 +338,6 @@ func bootstrapBucketEncryption(t *testing.T, api *s3api.Client, ctx context.Cont
 					ApplyServerSideEncryptionByDefault: &s3types.ServerSideEncryptionByDefault{
 						SSEAlgorithm: s3types.ServerSideEncryptionAes256,
 					},
-					BucketKeyEnabled: true,
```
simar7, line 341: I would keep this in, as it is another way for a bucket to be encrypted. As for the test, if you implement the suggestion I mentioned above, the test should be green for the cases when either `BucketKeyEnabled` and / or `ApplyServerSideEncryptionByDefault` is set.

Ideally they should be two different cases but I think we can leave that for now if you wish.

simar7 (changes requested):
thanks looks good, I left some feedback in the comments.

int-tt:
Thank you for feedback!
I fixed it.

simar7:
> Thank you for feedback!
> I fixed it.

Looks like there's a few checks that are red, could you address those?

simar7 (approved):
thanks!

PR #1088, opened 2022-11-29
feat(cloud): Enhance metadata support for rego policies

simar7:
Currently we don't support example metadata for Rego policies. This PR adds support for it.

This will be needed for those rules that currently exist in Go but are being ported over to Rego. We would need to bring the example help text over as well so that the AVD pages can be generated appropriately with the help texts. The help text is used as follows:
image

An example policy could look like the following (just for illustration):
```rego
# METADATA
# title: "RDS IAM Database Authentication Disabled"
# description: "Ensure IAM Database Authentication is enabled for RDS database instances to manage database access"
# scope: package
# schemas:
# - input: schema.input
# related_resources:
# - https://docs.aws.amazon.com/neptune/latest/userguide/iam-auth.html
# custom:
#   avd_id: AVD-AWS-0176
#   provider: aws
#   service: rds
#   severity: MEDIUM
#   short_code: enable-iam-auth
#   recommended_action: "Modify the PostgreSQL and MySQL type RDS instances to enable IAM database authentication."
#   input:
#     selector:
#     - type: cloud
#   cloud_formation:
#     good_examples:
#       - |
#         ---
#         Resources:
#         GoodExample:
#         Type: AWS::Logs::LogGroup
#         Properties:
#           KmsKeyId: "arn:aws:kms:us-west-2:111122223333:key/lambdalogging"
#           LogGroupName: "aws/lambda/goodExample"
#           RetentionInDays: 30
package builtin.aws.rds.aws0176

deny[res] {
	instance := input.aws.rds.instances[_]
	instance.engine.value == ["postgres", "mysql"][_]
	not instance.iamauthenabled.value
	res := result.new("Instance does not have IAM Authentication enabled", instance.iamauthenabled)
}

```

Observe the `cloud_formation` section in the metadata with the good examples. A similar one can be added for `terraform` as well.


pkg/rego/metadata.go
```diff
@@ -227,9 +231,44 @@ func (m *MetadataRetriever) updateMetadata(meta map[string]interface{}, metadata
...
+	if raw, ok := meta[schema]; ok {
+		sMap, ok = raw.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("failed to parse %s metadata: not an object", schema)
+		}
+	}
+
+	var em scan.EngineMetadata
+	if sMap["good_examples"] != nil {
+		em.GoodExamples = []string{fmt.Sprintf("%s", sMap["good_examples"])}
```
chen-keinan, line 258: maybe its good add a cast check , for example : 
```
val, ok := sMap["good_examples"].(string)
	if ok {
             em.GoodExamples = []string{fmt.Sprintf("%s",val)}
         }
```
otherwise this could happen incase of  `sMap["good_examples"] = 1`

`em.GoodExamples = []string{%!s(int=1)}`

pkg/rego/metadata.go
```diff
@@ -227,9 +231,44 @@ func (m *MetadataRetriever) updateMetadata(meta map[string]interface{}, metadata
...
+	}
+	if sMap["bad_examples"] != nil {
+		em.BadExamples = []string{fmt.Sprintf("%s", sMap["bad_examples"])}
```
chen-keinan, line 261: same as above

pkg/rego/metadata.go
```diff
@@ -227,9 +231,44 @@ func (m *MetadataRetriever) updateMetadata(meta map[string]interface{}, metadata
...
+	}
+	if sMap["links"] != nil {
+		em.Links = []string{fmt.Sprintf("%s", sMap["links"])}
```
chen-keinan, line 264: same as above

pkg/rego/metadata.go
```diff
@@ -227,9 +231,44 @@ func (m *MetadataRetriever) updateMetadata(meta map[string]interface{}, metadata
...
+	}
+	if sMap["remediation_markdown"] != nil {
+		em.RemediationMarkdown = fmt.Sprintf("%s", sMap["remediation_markdown"])
```
chen-keinan, line 267: same as above

pkg/rego/metadata.go
simar7: thanks, addressed.

cmd/avd_generator/main.go
```diff
@@ -53,35 +53,39 @@ func writeDocsFile(meta rules.RegisteredRule) {
...
+		if len(meta.Rule().Terraform.GoodExamples) > 0 || len(meta.Rule().Terraform.Links) > 0 {
+			tmpl, err := template.New("terraform").Parse(terraformMarkdownTemplate)
+			if err != nil {
+				fail("error occurred creating the template %v\n", err)
+			}
+			file, err := os.Create(filepath.Join(docpath, "Terraform.md"))
+			if err != nil {
+				fail("error occurred creating the Terraform file for %s", docpath)
+			}
+
```
chen-keinan, line 65:
```suggestion
			file, err := os.Create(filepath.Join(docpath, "Terraform.md"))
			if err != nil {
				fail("error occurred creating the Terraform file for %s", docpath)
			}
			defer func() {
				_ = file.Close()
			}()

```

cmd/avd_generator/main.go
```diff
@@ -53,35 +53,39 @@ func writeDocsFile(meta rules.RegisteredRule) {
...
-			fail("error occurred generating the document %v", err)
+		if len(meta.Rule().CloudFormation.GoodExamples) > 0 || len(meta.Rule().CloudFormation.Links) > 0 {
+			tmpl, err := template.New("cloudformation").Parse(cloudformationMarkdownTemplate)
+			if err != nil {
+				fail("error occurred creating the template %v\n", err)
+			}
+			file, err := os.Create(filepath.Join(docpath, "CloudFormation.md"))
+			if err != nil {
+				fail("error occurred creating the CloudFormation file for %s", docpath)
+			}
```
chen-keinan, line 82: same as above

chen-keinan:
LGTM 🚀   I have added a comment on AVD generator , it is always good to close the file handler to avoid exceeding the open handler limits

chen-keinan (dismissed):

cmd/avd_generator/main.go
simar7: yeah good point, ill add this in.

PR #1110, opened 2022-12-22
add SetInputSchema function to ConfigurableScanner

tamirkiviti13:
For custom scanners, this PR enables the option to set the rego scanner's json schema from outside

### DIFF
--- a/pkg/rego/scanner.go
+++ b/pkg/rego/scanner.go
@@ -106,6 +106,10 @@ func (s *Scanner) SetSkipRequiredCheck(_ bool) {
 	// NOTE: Skip required option not applicable for rego.
 }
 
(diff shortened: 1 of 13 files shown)


simar7:
thanks for the PR, but I don't understand how this will be used. Can you share an example?

pkg/rego/scanner.go
```diff
@@ -106,6 +106,10 @@ func (s *Scanner) SetSkipRequiredCheck(_ bool) {
 	// NOTE: Skip required option not applicable for rego.
 }
 
+func (s *Scanner) SetInputSchema(schema interface{}) {
+	s.inputSchema = schema
```
itaysk, line 110: wouldn't the inputSchema get overridden be the rego wrapper? i.e in L152?

PR #1111, opened 2022-12-22
Migrating RBAC Scanner to Kubernetes

sm171190:
In this change we have : 

1. migrated all internal rbac policies to kubernetes,
2.  removed the rbac scanner. This is because the resources Roles and ClusterRoles are essentially Kubernetes resources and hence should come under the scope of the Kubernetes scanner. 
3. we have ensured the unit tests are still functional.


avd_docs/kubernetes/general/AVD-KSV-0041/docs.md
```diff
@@ -0,0 +1,13 @@
+
```
simar7, line 1: So I see this was added (generated) but the corresponding RBAC docs are still present: https://github.com/sm171190/defsec/blob/sa-migrating-rbac-to-kubernetes-scanner/avd_docs/rbac/general/AVD-KSV-0041/docs.md


I assume there's no point in keeping RBAC docs, correct?

simar7 (changes requested):

simar7:
Another point is that you'd have to redo the IDs (for the newer policies) when this is merged https://github.com/aquasecurity/defsec/pull/1124

avd_docs/kubernetes/general/AVD-KSV-0041/docs.md
sm171190: That's correct. We'll have to remove the RBAC docs.

simar7:
Please update the IDs as we’ve merged https://github.com/aquasecurity/defsec/pull/1124

sm171190:
> Please update the IDs as we’ve merged #1124

@simar7 Have removed the rbac docs as they were duplicate and redundant. I don't think we need to change the IDs as PR #1124 has added cisbenchmarks policies which are indexed "KCV-" as opposed to general policies which are "KSV-". In fact, 
[https://github.com/aquasecurity/defsec/tree/master/avd_docs/kubernetes/general] has the last KSV-0040 and as part of this change we are continuing from KSV-0041 onwards. I think we should be good on this front.

simar7 (changes requested):
thanks the docs are looking good. just a quick note:

I think we need to change this to use the `kubernetes` schema going forwards for all `rbac` as input schema:
https://github.com/aquasecurity/defsec/blob/ad5cd7baeea3e2e15be8dc21eca6c0ced6545b96/pkg/rego/scanner.go#L121

so essentially:
```go
 types.SourceRbac:       schemas.Kubernetes, 
```

and you could comment this here (as it is done for "defsec" schema policies to use cloud instead):
https://github.com/aquasecurity/defsec/blob/418759b4dc97af25f30f32e0bd365be7984003a1/pkg/types/sources.go#L8
```go
SourceRbac       Source = "rbac" // deprecated - please use "kubernetes" instead
```

sm171190:
> thanks the docs are looking good. just a quick note:
> 
> I think we need to change this to use the `kubernetes` schema going forwards for all `rbac` as input schema:
> 
> https://github.com/aquasecurity/defsec/blob/ad5cd7baeea3e2e15be8dc21eca6c0ced6545b96/pkg/rego/scanner.go#L121
> 
> so essentially:
> 
> ```go
>  types.SourceRbac:       schemas.Kubernetes, 
> ```
> 
> and you could comment this here (as it is done for "defsec" schema policies to use cloud instead):
> 
> https://github.com/aquasecurity/defsec/blob/418759b4dc97af25f30f32e0bd365be7984003a1/pkg/types/sources.go#L8
> 
> ```go
> SourceRbac       Source = "rbac" // deprecated - please use "kubernetes" instead
> ```

@simar7 
Thanks for the comments. Have made the requested changes.

simar7:
Thanks for the changes but after reviewing them again and using them in Trivy, I feel we should first prepare a PR in Trivy that takes care of all RBAC removal related things.

This is because:
1. With this PR, defsec as a dependency breaks within Trivy. As seen below:
```
go: finding module for package github.com/aquasecurity/defsec/pkg/scanners/rbac
github.com/aquasecurity/trivy/pkg/fanal/handler/misconf imports
        github.com/aquasecurity/defsec/pkg/scanners/rbac: module github.com/aquasecurity/defsec@latest found (v0.82.6, replaced by /Users/simar/repos/defsec), but does not contain package github.com/aquasecurity/defsec/pkg/scanners/rbac
```
2. There's RBAC related filtering logic within Trivy that I assume will need to be removed (@chen-keinan or @josedonizetti could you advise?) https://github.com/aquasecurity/trivy/blob/9b0e9794cb94a2292c506afbff2bcd8695170558/pkg/k8s/report/report.go#L176

Once we have a PR ready in Trivy that uses this PR (as a defsec dependency), we should merge this PR first then the Trivy PR to avoid any breakage. 

Not doing so will not cause any breakage per se (as dependencies are pinned) but also will cause us to not be able to update version of defsec in Trivy until the Trivy PR (mentioned above) is merged.

sm171190:
@simar7 What you pointed out makes sense. I think @chen-keinan or @josedonizetti may be the correct folks to handle these changes on Trivy-side. Trivy  may need to start using the kubernetes scanner wherever we are using the rbac scanner curently because the kubernetes scanner now handles _all_ kubernetes resources (including rbac). But there may be other related changes on Trivy that I may not be aware of that may result because of this. I'll leave it to the guys more familiar with Trivy.  
@chen-keinan @josedonizetti Please suggest

chen-keinan:
@simar7 @sm171190 I'll do a quick test on `trivy` with this PR make sure it working and update you

sm171190:
> Thanks for the changes but after reviewing them again and using them in Trivy, I feel we should first prepare a PR in Trivy that takes care of all RBAC removal related things.
> 
> This is because:
> 
> 1. With this PR, defsec as a dependency breaks within Trivy. As seen below:
> 
> ```
> go: finding module for package github.com/aquasecurity/defsec/pkg/scanners/rbac
> github.com/aquasecurity/trivy/pkg/fanal/handler/misconf imports
>         github.com/aquasecurity/defsec/pkg/scanners/rbac: module github.com/aquasecurity/defsec@latest found (v0.82.6, replaced by /Users/simar/repos/defsec), but does not contain package github.com/aquasecurity/defsec/pkg/scanners/rbac
> ```
> 
> 2. There's RBAC related filtering logic within Trivy that I assume will need to be removed (@chen-keinan or @josedonizetti could you advise?) https://github.com/aquasecurity/trivy/blob/9b0e9794cb94a2292c506afbff2bcd8695170558/pkg/k8s/report/report.go#L176
> 
> Once we have a PR ready in Trivy that uses this PR (as a defsec dependency), we should merge this PR first then the Trivy PR to avoid any breakage.
> 
> Not doing so will not cause any breakage per se (as dependencies are pinned) but also will cause us to not be able to update version of defsec in Trivy until the Trivy PR (mentioned above) is merged.
> 
> @sm171190 - does it make sense? If not let's get on a call to discuss. As for the Trivy PR itself, I'm happy to review for you if you can send us one.

@simar7 The Trivy [PR](https://github.com/aquasecurity/trivy/pull/3432) is ready . Can you please merge this PR. I'll ask @chen-keinan to merge the Trivy PR next. We will then haave 1 more PR on Trivy-Operator which will also incorporate the same changes.

chen-keinan:
@sm171190 @simar7 I think the defsec PR shoylb be merged 1st and then we can update trivy PR , with latest of Defsec

simar7 (approved):

PR #1138, opened 2023-01-17
feat(terraform): hashicorp/external provider

lyoung-confluent:
Adds a provider/parser for [hashicorp/external](https://registry.terraform.io/providers/hashicorp/external/latest/docs) which can be used to execute external programs, exposing the results as values within the template. I did not introduce any associated rules, but an example rule could restricting which external programs are allowed based on a rego rule for the `program` attribute.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aquasecurity/defsec/pkg/providers/azure"
 	"github.com/aquasecurity/defsec/pkg/providers/cloudstack"
 	"github.com/aquasecurity/defsec/pkg/providers/digitalocean"
+	"github.com/aquasecurity/defsec/pkg/providers/external"
 	"github.com/aquasecurity/defsec/pkg/providers/github"
 	"github.com/aquasecurity/defsec/pkg/providers/google"
 	"github.com/aquasecurity/defsec/pkg/providers/kubernetes"
@@ -20,6 +21,7 @@ type State struct {
 	Azure        azure.Azure
 	CloudStack   cloudstack.CloudStack
 	DigitalOcean digitalocean.DigitalOcean
+	External     external.External
 	GitHub       github.GitHub
 	Google       google.Google
 	Kubernetes   kubernetes.Kubernetes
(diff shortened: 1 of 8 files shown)


simar7:
thanks @lyoung-confluent for your contribution! This is a cool idea. Could you please sign the CLA first?

Once you have that, could we explore some example rego policies that could be used with this adapter? Would love to learn more on what you have in mind.

lyoung-confluent:
@simar7 I have signed the CLA, the rego policy I was starting out with was just banning the use of `external` altogether as the legitimate use-cases for it are pretty rare (it's more commonly abused to get RCE via `terraform plan`), but I wanted to introduce it as part of the rego input so if there were legitimate use-cases they could be checked against with a more specific rule

pkg/state/state.go
```diff
@@ -20,6 +21,7 @@ type State struct {
 	Azure        azure.Azure
 	CloudStack   cloudstack.CloudStack
 	DigitalOcean digitalocean.DigitalOcean
+	External     external.External
```
simar7, line 24: just a nit: but maybe the provider name could be more descriptive to indicate that it's only terraform related? A suggestion: `TerraformExternal`

It does feel a little odd to include this here TBH, but then again the only "providers" we have so far are "service related" like AWS, Azure, etc.

What do you think?

lyoung-confluent: Yeah I had a similar uncertainty. However it is actually a provider, not something built-in to terraform stdlib: https://registry.terraform.io/providers/hashicorp/external/latest/docs 
So if anything different it should be `HashicorpExternal` based on the provider namespace?

I don't have a strong opinion on what it should be, though I will note I am working on another PR that does add internal Terraform structures (imported modules, provisioners, etc) so they can be queried/matched agains with rego if that changes your opinon on how this one should be named. I was thinking that one will be just `Terraform` here with child resources like `Modules` and `Provisioners` within.

simar7: > I don't have a strong opinion on what it should be, though I will note I am working on another PR that does add internal Terraform structures (imported modules, provisioners, etc) so they can be queried/matched agains with rego if that changes your opinon on how this one should be named. I was thinking that one will be just `Terraform` here with child resources like `Modules` and `Provisioners` within.

Yeah that sounds good to me - what about a `Hashicorp` package that includes this provider and the other `Terraform` related things (child resources like modules and provisioners as you mentioned).

PR #1146, opened 2023-01-23
fix(rules) updated logic for ec2 nacl rule

giorod3:
Addresses: https://github.com/aquasecurity/trivy/issues/3283

### DIFF
--- a/internal/rules/aws/ec2/no_excessive_port_access.go
+++ b/internal/rules/aws/ec2/no_excessive_port_access.go
@@ -15,7 +15,7 @@ var CheckNoExcessivePortAccess = rules.Register(
 		Provider:    providers.AWSProvider,
 		Service:     "ec2",
 		ShortCode:   "no-excessive-port-access",
-		Summary:     "An ingress Network ACL rule allows ALL ports.",
+		Summary:     "An egress Network ACL rule allows ALL ports.",
 		Impact:      "All ports exposed for egressing data",
 		Resolution:  "Set specific allowed ports",
 		Explanation: `Ensure access to specific required ports is allowed, and nothing else.`,
@@ -39,7 +39,7 @@ var CheckNoExcessivePortAccess = rules.Register(
 	func(s *state.State) (results scan.Results) {
 		for _, acl := range s.AWS.EC2.NetworkACLs {
 			for _, rule := range acl.Rules {
-				if rule.Protocol.EqualTo("-1") || rule.Protocol.EqualTo("all") {
+				if rule.Type.EqualTo("egress") && rule.Action.EqualTo("allow") && rule.Protocol.EqualTo("-1") || rule.Protocol.EqualTo("all") {
 					results.Add(
 						"Network ACL rule allows access using ALL ports.",
 						rule.Protocol,
--- a/internal/rules/aws/ec2/no_excessive_port_access.go
+++ b/internal/rules/aws/ec2/no_excessive_port_access.go
@@ -15,8 +15,8 @@ var CheckNoExcessivePortAccess = rules.Register(
 		Provider:    providers.AWSProvider,
 		Service:     "ec2",
 		ShortCode:   "no-excessive-port-access",
-		Summary:     "An egress Network ACL rule allows ALL ports.",
-		Impact:      "All ports exposed for egressing data",
+		Summary:     "An Network ACL rule allows ALL ports.",
+		Impact:      "All ports exposed for ingressing/egressing data",
 		Resolution:  "Set specific allowed ports",
 		Explanation: `Ensure access to specific required ports is allowed, and nothing else.`,
 		Links: []string{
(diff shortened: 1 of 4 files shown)


internal/rules/aws/ec2/no_excessive_port_access.go
```diff
@@ -15,7 +15,7 @@ var CheckNoExcessivePortAccess = rules.Register(
 		Provider:    providers.AWSProvider,
 		Service:     "ec2",
 		ShortCode:   "no-excessive-port-access",
-		Summary:     "An ingress Network ACL rule allows ALL ports.",
+		Summary:     "An egress Network ACL rule allows ALL ports.",
```
knqyf263, line 18: This check is supposed to be about ingress. I think changing to egress would be a completely different rule. Am I missing something?

internal/rules/aws/ec2/no_excessive_port_access.go
```diff
@@ -39,7 +39,7 @@ var CheckNoExcessivePortAccess = rules.Register(
 	func(s *state.State) (results scan.Results) {
 		for _, acl := range s.AWS.EC2.NetworkACLs {
 			for _, rule := range acl.Rules {
-				if rule.Protocol.EqualTo("-1") || rule.Protocol.EqualTo("all") {
+				if rule.Type.EqualTo("egress") && rule.Action.EqualTo("allow") && rule.Protocol.EqualTo("-1") || rule.Protocol.EqualTo("all") {
```
knqyf263, line 42: I thought we should confirm `rule_action` is `allow`.

giorod3: we are confirming the rule_action is allow. I added egress to align with the check.

internal/rules/aws/ec2/no_excessive_port_access.go
giorod3: If you look at impact, it is referring to egress. Egress is more important when talking about NACLs since thats where data leakage would come from, and in CSPM we have the same check that is looking specifically for egress. I can change it back to ingress but the logic in the plugin was not checking ingress/egress anyways. 

what do you think?

knqyf263: I guess the impact is wrong as the severity is critical. I agree that egress is also important, but I don't think it is a critical issue, while ingress is critical. I'm not sure how the check was added originally. But it is a good idea to get aligned with our CSPM.
We can also hear @simar7's thoughts.

internal/rules/aws/ec2/no_excessive_port_access.go
knqyf263: Yeah, I saw it. We should discuss ingress/egress. I'll mark it as resolved.

internal/rules/aws/ec2/no_excessive_port_access.go
giorod3: here is the cloudsploit check for reference: https://github.com/aquasecurity/cloudsploit/blob/master/plugins/aws/ec2/networkAclOutboundTraffic.js

knqyf263: Interesting. I thought ingress was more important than egress, but CloudSploit thinks the opposite as they have no check for nacl ingress. You can decide then. I'm ok with the current change.

simar7: I think you both are right - so there are two things I found:

1. We have a rule for ingress: https://github.com/aquasecurity/defsec/blob/master/internal/rules/aws/ec2/no_public_ingress_acl.go
2. As for severity, Ingress to me certainly sounds CRITICAL as Teppei mentioned. But for egress, EC2 instances start wide open by default to my knowledge (no restrictions on egress traffic). Based on that, maybe we could lower the severity to HIGH or MEDIUM, but it is subjective as there isn't any guideline or spec to follow.

knqyf263: Thanks, Simar. According to the rule Simar found, I guess this check is also for ingress originally.

>1. We have a rule for ingress: >https://github.com/aquasecurity/defsec/blob/master/internal/rules/aws/ec2/no_public_ingress_acl.go

But egress is also important as @giorod3 said. I agree we should have both. There are two options now.

1. Apply this check to both ingress and egress as it is now
    - In this case, we just add `rule.Action.EqualTo("allow") &&` in order to fix the original bug.
2. Split this check into two checks for having different severities
    - ingress: critical
    - egress: medium or high

@giorod3 should be the most familiar with CSPM. You can decide!

giorod3:
@simar7 @knqyf263 from my understanding its just about excessive ports not necessarily about ingress/egress. for now I have just reverted the logic that checks for egress but I think we should revisit this when we are migrating to rego as I don't think its a relevent check when we have specific checks for both egress and ingress. I have also changed the impact messaging.

simar7 (approved):
lgtm 🚢

PR #1149, opened 2023-01-24
fix(rego): metadata for rego results

### LINKED ISSUE #1148: bug: incorrect/broken path on rego findings in remote terraform modules
**Describe the bug**
When (custom) rego findings occur in remote terraform module, the reported file path is incorrect and fails to render in the output:
```
  Failed to render code: failed to read file from result filesystem ("/var/folders/z7/lmc4dbqj43x6gxhhw38__hcr0000gq/T/.aqua/cache/61225a9ded8f54e2b6f7871399933e93"): open /var/folders/z7/lmc4dbqj43x6gxhhw38__hcr0000gq/T/.aqua/cache/61225a9ded8f54e2b6f7871399933e93/github.com/org/repo?ref=master/vuln.tf: no such file or directory
```
This issue does not occur for non-rego findings.

**To Reproduce**
Create a repository with a remote Terraform module, ex:
```hcl
module "remote" {
  source = "github.com/org/repo?ref=master"
}
```
Populate the remote repository with some vulnerable terraform:
```hcl
resource "aws_s3_bucket" "demo" {
  bucket = "insecure"
}
```
When scanning with the built-in Golang rules, it will render correctly:
```
Result #5 HIGH No public access block so not blocking public acls
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  github.com/org/repo?ref=master/vuln.tf:11-13
   via vuln.tf:24-27 (module.remote)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   11    resource "aws_s3_bucket" "demo" {
   12      bucket = "insecure"
   13    }
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
          ID aws-s3-block-public-acls
      Impact PUT calls with public ACLs specified can make objects public
  Resolution Enable blocking any PUT calls with a public ACL specified

  More Information
  - https://aquasecurity.github.io/tfsec/latest/checks/aws/s3/block-public-acls/
  - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_public_access_block#block_public_acls
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```
However if a custom rego rule matches the same file, an error occurs and the context (path to the vulnerable tf) is lost:
```
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  ../../../../../../github.com/org/repo?ref=master/vuln.tf:6-8
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  Failed to render code: failed to read file from result filesystem ("/var/folders/z7/lmc4dbqj43x6gxhhw38__hcr0000gq/T/.aqua/cache/61225a9ded8f54e2b6f7871399933e93"): open /var/folders/z7/lmc4dbqj43x6gxhhw38__hcr0000gq/T/.aqua/cache/61225a9ded8f54e2b6f7871399933e93/github.com/org/repo?ref=master/vuln.tf: no such file or directory
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  Rego Package custom.demo
     Rego Rule deny
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

**Expected behavior**
The rego result contains the nested path to reach the vulnerable code and outputs the correct snippet.

lyoung-confluent:
Fixes #1148 by adding the `Parent` and `SourcePrefix` fields to the generated rego input/metadata, and using those fields to reconstruct the original metadata when results are matched.


pkg/scan/result.go
```diff
@@ -281,12 +281,16 @@ func (r *Results) SetSourceAndFilesystem(source string, f fs.FS, logicalSource b
...
 		}
+		parent := m.Parent()
 		switch {
 		case m.IsExplicit():
 			m = defsecTypes.NewExplicitMetadata(newrng, m.Reference())
 		default:
 			m = defsecTypes.NewMetadata(newrng, m.Reference())
 		}
+		if parent != nil {
+			m.SetParentPtr(parent)
```
lyoung-confluent, line 292: I'm unsure if this is sufficient/don't fully understand the use-case for this code. It may be that the metadata of the parent should also be cloned/overwritten here.

simar7:
Thanks for the PR, will take a look.

pkg/types/metadata.go
```diff
@@ -77,15 +77,20 @@ func (m *Metadata) UnmarshalJSON(data []byte) error {
...
+		"sourceprefix": m.Range().GetSourcePrefix(),
+		"managed":      m.isManaged,
+		"explicit":     m.isExplicit,
+		"fskey":        CreateFSKey(m.Range().GetFS()),
+		"resource":     m.Reference(),
+	}
+	if m.parent != nil {
+		input["parent"] = m.parent.ToRego()
 	}
+	return input
```
simar7, line 93: I see you've refactored this and used it in all the `ToRego()` calls that are elsewhere, which is fantastic. Could you add a basic unit test that exercises this? 

So basically `metadata_test.go` to go along with this that can exercise the various calls. I would create a `NewMetadata` and then simply call `ToRego` on it to observe and assert the behaviour.

simar7: That way you won't have to test every single type (string, int, etc.) individually.

lyoung-confluent: added a test that exercises the non-parent and parent use-cases

pkg/rego/custom.go
```diff
@@ -28,24 +28,37 @@ func init() {
 	)
 }
 
-func createResult(_ rego.BuiltinContext, msg, cause *ast.Term) (*ast.Term, error) {
+func createResult(ctx rego.BuiltinContext, msg, cause *ast.Term) (*ast.Term, error) {
```
simar7, line 31: Looks like this is currently untested, can we add a unit test for it?

pkg/scan/result.go
simar7: What happens if you don't set the parent? Maybe we can try to add a test for this to observe and document the expected behaviour.

lyoung-confluent: If you don't set the parent the output will be missing the correct context of how the finding was reached, ex (from `tfsec` formatter):
```
──────────────────────────────────────────────────
  github.com/org/repo.git?ref=master/pwn.tf:11-13
──────────────────────────────────────────────────
```
Instead of the correct output:
```
──────────────────────────────────────────────────
  github.com/org/repo.git?ref=master/pwn.tf:11-13
   via test.tf:14-17 (module.example)
──────────────────────────────────────────────────
```
So the adding of the parent is necessary (one of the core features of this PR), it's more that this method (`SetSourceAndFilesystem`) is replacing the source/filesystem of the metadata in the results which maybe should include the parent metadata as well

simar7: Right, I understand - I think there's value in testing this in isolation as well. At the moment the only tests we seem to have exercise this indirectly via the actual scanners, for example a call to: https://github.com/aquasecurity/defsec/blob/3ba2da0e8a11a7eff2e2f73fdc44d139791ac15d/pkg/scanners/dockerfile/scanner_test.go#L60 - will result in `SetSourceAnFilesystem` getting called. 

The other way to test this would be to add this assertion to each of the scanners, unless you've already done that in this PR, which I might have missed. 


What do you think?

simar7:
hi @lyoung-confluent - are you still looking into this? Would be nice to have it in.

simar7 (approved):
I added some more tests, but lgtm. thanks again for the contribution.

PR #1151, opened 2023-01-26
fix(aws/sam): fix HttpApi access logging rule

pjungermann:
`AWS::Serverless::HttpApi` uses `AccessLogSettings` instead of `AccessLogSetting`.

See
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-httpapi.html#sam-httpapi-accesslogsettings


pjungermann:
not stale

simar7 (approved):
thanks, lgtm. sorry for the delay in reviewing.

PR #1156, opened 2023-01-30
fix(compliance): remove unneeded hyphen

knqyf263:
The output is broken now.
image


chen-keinan (approved):

knqyf263:
This is a tiny fix. I don't think I have to wait for reviews from code owners.

PR #1159, opened 2023-01-31
chore: kubelet wrong check mapping

chen-keinan:
kubelet wrong check mapping


simar7:
The renaming is fine but upon running `make docs` (we should probably include this as a GitHub action step) it generated a diff, which meant at some point the new docs might not have been committed in a previous PR.

I've generated and committed them. When the test pass, I'll merge it in and release a new defsec.

simar7 (approved):

PR #1188, opened 2023-02-17
feat(terraform) adding trivy:ignore for tf scans to align with trivy

giorod3:
Aligning with trivy to allow inline ignores using trivy:ignore.


pkg/scanners/terraform/scanner_test.go
```diff
@@ -66,6 +66,24 @@ resource "something" "else" {}
...
+	}
+	results := scanWithOptions(t, `
+//trivy:ignore:altid
+resource "something" "else" {}
+`, options...)
+	require.Len(t, results.GetFailed(), 0)
+	require.Len(t, results.GetIgnored(), 1)
+
+}
+
```
simar7, line 86: As most of the other inputs are the same, can we make this into a table driven test instead? So something like:

testCases := []struct{
name string
inputOptions string}{
{ name: "tfsec", inputOptions: "..."}}

then you can iterate over them as inputs:

for _, tc := range testCases {
t.Run(...)
}

test/trivy_ignore_test.go
```diff
@@ -0,0 +1,393 @@
+package test
```
simar7, line 1: Likewise with this file as well. `ignore_test.go` could include all the ignores, trivy and tfsec as test inputs for each of the existing tests within `ignore_test.go`.

test/trivy_ignore_test.go
```diff
@@ -0,0 +1,393 @@
...
+	"github.com/aquasecurity/defsec/pkg/severity"
+
+	"github.com/aquasecurity/defsec/pkg/scan"
+
+	"github.com/aquasecurity/defsec/internal/rules"
+
+	"github.com/aquasecurity/defsec/pkg/providers"
+
+	"github.com/stretchr/testify/assert"
+)
```
simar7, line 18: lint:

```suggestion
	"fmt"
	"testing"

	"github.com/aquasecurity/defsec/internal/rules"
	"github.com/aquasecurity/defsec/pkg/providers"
	"github.com/aquasecurity/defsec/pkg/scan"
	"github.com/aquasecurity/defsec/pkg/severity"
	"github.com/aquasecurity/defsec/pkg/terraform"
	"github.com/stretchr/testify/assert"
)
```

If you're IntelliJ, you can setup `gofmt` to run as a file watcher ever time you make a change.

giorod3:
@simar7 I made the changes we discussed. please review and merge if all looks good

simar7 (approved):

PR #1190, opened 2023-02-19
feat(compliance): add special ids for summarizing vulnerabilities

knqyf263:
I've added two special IDs for filtering vulnerabilities by severity in compliance reports.
- VULN-CRITICAL
- VULN-HIGH

It allows enriching Docker CIS Benchmark. Secrets will be added in another PR.


chen-keinan (approved):

chen-keinan:
@knqyf263 A new package need to be published , so trivy can pick up this spec

PR #1191, opened 2023-02-19
feat(rego): accept fs.FS for data

knqyf263:
Add a new option to pass the data file system so that the file system can be abstract.
https://github.com/aquasecurity/trivy/pull/3578

### DIFF
--- a/pkg/rego/load.go
+++ b/pkg/rego/load.go
@@ -156,7 +156,7 @@ func (s *Scanner) LoadPolicies(loadEmbedded bool, srcFS fs.FS, paths []string, r
 	for namespace := range uniq {
 		namespaces = append(namespaces, namespace)
 	}
-	store, err := initStore(s.dataDirs, namespaces)
+	store, err := initStore(s.dataDirs, namespaces, s.data)
 	if err != nil {
 		return fmt.Errorf("unable to load data: %w", err)
 	}
--- a/pkg/rego/load.go
+++ b/pkg/rego/load.go
@@ -156,7 +156,13 @@ func (s *Scanner) LoadPolicies(loadEmbedded bool, srcFS fs.FS, paths []string, r
 	for namespace := range uniq {
 		namespaces = append(namespaces, namespace)
 	}
-	store, err := initStore(s.dataFS, s.dataDirs, namespaces)
+
+	dataFS := srcFS
+	if s.dataFS != nil {
+		s.debug.Log("Overriding filesystem for data!")
+		dataFS = s.dataFS
+	}
+	store, err := initStore(dataFS, s.dataDirs, namespaces)
 	if err != nil {
 		return fmt.Errorf("unable to load data: %w", err)
 	}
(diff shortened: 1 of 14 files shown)


chen-keinan:
> 
> @chen-keinan What if passing data directly rather than creating a temp file? I've created a PoC. Would you take a look?

since trivy already support accepting  rego data, maybe it better to enhance it to accept `in memory FS` rather than real fs and use the current logic exist in trivy , probably will be much simpler, wdyt ?

knqyf263:
`memfs` is supported for input, but not for data. As far as I understand, OPA cannot mix real files and Go structs.
https://github.com/open-policy-agent/opa/blob/c0360f1af571c1087be574a361d06b8961b1fb93/loader/loader.go#L121-L127

We need to load files, create a memory filesystem, and add the Go struct there, then we can pass it to `FileLoader`. Passing `map[string]any` to `loader.Result` looks simpler to me. But I might be wrong as I'm not following Rego updates recently.

Also, I think data is likely passed as a Go struct in the OPA world. It doesn't have to be a file.
https://github.com/open-policy-agent/opa/blob/c0360f1af571c1087be574a361d06b8961b1fb93/rego/example_test.go#L324-L335

knqyf263:
I've updated PR to accept `fs.FS` for `data` according to Chen's suggestion.
https://github.com/aquasecurity/defsec/pull/1191/commits/c2f50401014ecf3e0f892e1e88ca8c33c2661538

pkg/scanners/helm/scanner.go
```diff
@@ -114,6 +114,9 @@ func (s *Scanner) SetPolicyFilesystem(policyFS fs.FS) {
 	s.policyFS = policyFS
 }
 
+func (s *Scanner) SetDataFilesystem(policyFS fs.FS) {
```
chen-keinan, line 117:
```suggestion
func (s *Scanner) SetDataFilesystem(dataFS fs.FS) {
```

pkg/rego/load.go
```diff
@@ -156,7 +156,7 @@ func (s *Scanner) LoadPolicies(loadEmbedded bool, srcFS fs.FS, paths []string, r
 	for namespace := range uniq {
 		namespaces = append(namespaces, namespace)
 	}
-	store, err := initStore(s.dataDirs, namespaces)
+	store, err := initStore(s.dataFS, s.dataDirs, namespaces)
```
chen-keinan, line 159: maybe `dataDirs` can be removed

knqyf263: After removing `dataDirs` locally, I changed my mind. `dataFS` possibly contains irrelevant files and dirs. We might want to filter out those files/dirs. I keep `dataDirs` for the reason.

pkg/scanners/helm/scanner.go
knqyf263: Done
https://github.com/aquasecurity/defsec/pull/1191/commits/082cb0d7b9b9c2f30180ca2a35d229472ff8596a

chen-keinan:
lgtm

chen-keinan (approved):

knqyf263:
@simar7 @giorod3 Could you take a look? We just want to load the OPA data from fs.FS.

simar7:
@knqyf263 the idea seems fine to me but how will we expose this to the callers? More specifically, would any trivy flags would need to be changed OR internally instead of creating temp files, the caller would pass along something that implements fs.FS?

chen-keinan (approved):

simar7 (approved):
lgtm

PR #1202, opened 2023-02-25
fix: output correct path for nested terraform modules (SARIF)

### LINKED ISSUE #1147: bug: SARIF outputs wrong path for nested local terraform modules
**Describe the bug**

When using locally resolved modules, the SARIF output wrongly attaches all issues to the root module import.

My hunch is, that it is a regression by https://github.com/aquasecurity/defsec/issues/888, as it introduced [this](https://github.com/aquasecurity/defsec/blob/046ee90ace59c3d87dab96889a87b636fdb205ef/pkg/formatters/sarif.go#L41) line. Due to the `Root()` call, all issues are resolved to the main module import.

**To Reproduce**

`tfsec --format sarif` against the provided example. The found issues are all with the same line numbers at `main.tf`.

**Expected behavior**

The issues should actually be linked to `modules/something/main.tf` (see example below for context)

**Output of your tfsec command with --debug flag**
Not applicable

**System Info**
Not applicable

**Example Code**

`main.tf`
```terraform
module "main" {
  source = "./modules/something"
}
```

`modules/something/main.tf` (just a random rule to provoke an issue)
```terraform
resource "aws_security_group_rule" "my-rule" {
  type        = "ingress"
  cidr_blocks = ["0.0.0.0/0"]
}
```

**Additional context**
Not applicable

vanesasejdiu:
closes #1147

### DIFF
--- a/pkg/formatters/sarif.go
+++ b/pkg/formatters/sarif.go
@@ -38,9 +38,7 @@ func outputSARIF(b ConfigurableFormatter, results scan.Results) error {
 			rule.WithHelpURI(links[0])
 		}
 
-		metadata := res.Metadata().Root()
-
-		rng := metadata.Range()
+		rng := res.Metadata().Range()
 		message := sarif.NewTextMessage(res.Description())
 		region := sarif.NewSimpleRegion(rng.GetStartLine(), rng.GetEndLine())
 		var level string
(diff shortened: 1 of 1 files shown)


giorod3 (approved):

giorod3:
@simar7 can you review and merge this as well

simar7:
Can we add a small test case (probably the input that the user gave in the https://github.com/aquasecurity/defsec/issues/1147 should work) so that we can test this and guard any future regressions?

simar7:
> Can we add a small test case (probably the input that the user gave in the #1147 should work) so that we can test this and guard any future regressions?

I've created a new issue to track adding this test as I do want to get this fix in the next release.

kernle32dll:
Duh, I don't know why GitHub did not send me a notification for this. Thanks a lot for fixing! Can confirm its working for us now.

PR #1210, opened 2023-02-27
feat(compliance): add special IDs for secrets

knqyf263:
Same as vulnerabilities, but for secrets.
https://github.com/aquasecurity/defsec/pull/1190


rules/specs/compliance/docker-cis.yaml
```diff
@@ -61,6 +61,7 @@ spec:
       name: Ensure secrets are not stored in Dockerfiles (Manual)
```
chen-keinan, line 61: remove `(Manual)`

chen-keinan (approved):

PR #1238, opened 2023-03-15
Added new check - object lock configuration.

VikrantCGaikwad:
Created new PR for new check object lock configuration...also resolved 1 issue in pkg/scanners/azure/functions/utc_now_test.go where utc time was not considered while setting expected time resulting in failed test

### DIFF
--- a/internal/adapters/cloudformation/aws/s3/bucket.go
+++ b/internal/adapters/cloudformation/aws/s3/bucket.go
@@ -27,8 +27,9 @@ func getBuckets(cfFile parser.FileContext) []s3.Bucket {
 				Enabled:   hasVersioning(r),
 				MFADelete: defsecTypes.BoolUnresolvable(r.Metadata()),
 			},
-			Logging: getLogging(r),
-			ACL:     convertAclValue(r.GetStringProperty("AccessControl", "private")),
+			Logging:                 getLogging(r),
+			ACL:                     convertAclValue(r.GetStringProperty("AccessControl", "private")),
+			ObjectLockConfiguration: getObjectLockConfiguration(r, cfFile),
 		}
 
 		buckets = append(buckets, s3b)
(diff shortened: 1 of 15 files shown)


rules/cloud/policies/aws/s3/enable_versioning.go
```diff
@@ -41,7 +41,8 @@ With versioning you can recover more easily from both unintended user actions an
 	},
 	func(s *state.State) (results scan.Results) {
 		for _, bucket := range s.AWS.S3.Buckets {
-			if !bucket.Versioning.Enabled.IsTrue() {
+			// Versioning check changed as Object lock can be configured on versioned buckets only
+			if !bucket.Versioning.Enabled.IsTrue() && !bucket.ObjectLockConfiguration.Enabled.IsTrue() {
```
simar7, line 45: Sorry but I don't follow, why does this logic needs to be changed? Your comment mentions that Object lock can only be configured on versioned buckets only but this check only checks for versioning to be enabled (or otherwise) for buckets.

rules/cloud/policies/aws/s3/enable_object_lock_cofiguration.cf.go
```diff
@@ -0,0 +1,32 @@
...
+    Properties:
+      ObjectLockConfiguration:
+        - ObjectLockEnabled: "Enabled"
+            ObjectLockRule:
+               	Days: 10
+ 				Mode: "COMPLIANCE"
+  				Years: 0
+    Type: AWS::S3::Bucket
+`,
+}
```
simar7, line 16:
```suggestion
var cloudFormationEnableObjectLockConfigurationGoodExamples = []string{
	`
Resources:
  GoodExample:
    Properties:
      ObjectLockConfiguration:
        - ObjectLockEnabled: "Enabled"
            ObjectLockRule:
               	Days: 10
 		Mode: "COMPLIANCE"
  		Years: 0
    Type: AWS::S3::Bucket
`,
}
```

internal/adapters/cloudformation/aws/s3/bucket.go
```diff
@@ -112,3 +113,17 @@ func getEncryption(r *parser.Resource, _ parser.FileContext) s3.Encryption {
 
 	return encryption
 }
+
+func getObjectLockConfiguration(r *parser.Resource, _ parser.FileContext) s3.ObjectLockConfiguration {
+	objectlockconfigs := s3.ObjectLockConfiguration{
+		Metadata: r.Metadata(),
+		Enabled:  defsecTypes.BoolDefault(false, r.Metadata()),
+		// DefaultRetention: defsecTypes.StringDefault("", r.Metadata()),
```
simar7, line 121:
```suggestion
```
don't think we check this so better to remove it altogether.

rules/cloud/policies/aws/s3/enable_versioning.go
VikrantCGaikwad: Screenshot from 2023-03-16 11-08-58
this is ticket description that needed to be implemented...as per point iii have added this condition

giorod3: Hi @VikrantCGaikwad if Object Lock is enabled, Versioning must be enabled. But if Versioning is enabled, Object Lock does not need to be enabled. This change will break our functionality.

VikrantCGaikwad: Hi @giorod3 @simar7 ...
if !bucket.Versioning.Enabled.IsTrue() && !bucket.ObjectLockConfiguration.Enabled.IsTrue() {
If this condition is true then we are seeting only Versioning not Object Lock Configuration...
**But if Versioning is enabled, Object Lock does not need to be enabled**  we are not setting this anywhere

PR #1244, opened 2023-03-24
fix the issue: outdated_tls_policy

SanaaYousaf:


SanaaYousaf:
this commit fixes #3007 https://github.com/aquasecurity/trivy/issues/3007

internal/adapters/cloudformation/aws/elb/loadbalancer.go
```diff
@@ -33,7 +33,7 @@ func getListeners(lbr *parser.Resource, ctx parser.FileContext) (listeners []elb
 			listener := elb.Listener{
 				Metadata:       r.Metadata(),
 				Protocol:       r.GetStringProperty("Protocol", "HTTP"),
-				TLSPolicy:      r.GetStringProperty("SslPolicy", "ELBSecurityPolicy-2016-08"),
+				TLSPolicy:      r.GetStringProperty("SslPolicy", ""),
```
simar7, line 36: Is there a failing test that we would need to update after this change? If not, we should add a test for it.

SanaaYousaf: No, there is no any test file for cloudformation adapter.

simar7: OK can we add a simple unit test in that case? The fix is straight forward but having a unit test will guard us against regressions in the future.

SanaaYousaf: I did not write unit test for cloudformation file before, so i have no idea how to test it..

simar7: OK looks like CloudFormation could be better tested. I'll open a separate issue for it as it seems there's no test in place at the moment. 

For your reference, we test the terraform equivalent as so: https://github.com/aquasecurity/defsec/blob/master/internal/adapters/terraform/aws/elb/adapt_test.go

simar7: Issue opened here: https://github.com/aquasecurity/defsec/issues/1257

simar7 (dismissed):

simar7 (approved):

PR #1246, opened 2023-03-25
Add check missing `microdnf clean all`

atombrella:
As an alternative to dnf, there's `microdnf` which comes with fewer options, but otherwise identical syntax for basic operations. Perhaps it could be merged with the existing check? It's mostly just copy/paste from the two dnf-clean-all files.

missing `--nodocs` for both `microdnf install` and `dnf install` is another good candidate for a valid check. Who needs docs in Docker containers? :D I could volunteer to add this as well.

`setopt=install_weak_deps=0` is what I guess equivalent to `apt` `--no-install-recommends`. There's no existing check for this.


simar7 (approved):
thanks this looks good to me, having another check for now will suffice.

simar7:
> As an alternative to dnf, there's `microdnf` which comes with fewer options, but otherwise identical syntax for basic operations. Perhaps it could be merged with the existing check? It's mostly just copy/paste from the two dnf-clean-all files.
> 

for sure, if you'd like to refactor some common code, would be happy to review it.

> missing `--nodocs` for both `microdnf install` and `dnf install` is another good candidate for a valid check. Who needs docs in Docker containers? :D I could volunteer to add this as well.
> 

sure sounds good!

> `setopt=install_weak_deps=0` is what I guess equivalent to `apt` `--no-install-recommends`. There's no existing check for this.

PR #1258, opened 2023-03-28
draft: add trivy ignore in cf files

vanesasejdiu:


pkg/scanners/cloudformation/test/examples/ignores/bucket_with_ignores.yaml
```diff
@@ -16,7 +16,7 @@ Resources:
         Ref: BucketName
       PublicAccessBlockConfiguration:
         BlockPublicAcls: false
-        BlockPublicPolicy: false # cfsec:ignore:AVD-AWS-0087
+        BlockPublicPolicy: false # trivy:ignore:AVD-AWS-0087
```
simar7, line 19: Does this mean we will no longer support `cfsec` prefixes? If so, that'll be backwards incompatible. We should support both (`trivy` and `cfsec`) at this time.

PR #1259, opened 2023-03-29
filter iam pass role in policy document to prevent privilege escalations

realwebdev:
prevent privilege escalation.


simar7:
can you run `make docs` and add the generated files in this commit? also could you add a simple unit test for this rule alongside it?

realwebdev:
> can you run `make docs` and add the generated files in this commit? also could you add a simple unit test for this rule alongside it?

I have added these.

PR #1266, opened 2023-04-04
WIP Add check for missing dnf install --nodcs

atombrella:
It's Work in Progress; it's better to go here to ask for feedback and help. I'm currently struggling a bit with the tests.

I don't know how much of a difference the flag makes, I haven't dug deep.


atombrella:
> hey @atombrella - the tests for this could be similar to the other PR you worked on #1270? Or was there something else you needed help with?

Well, the first test fails, and I think the regex is the culprit. I'm trying to re-work a bit, and hope to push something in the weekend :)

PR #1267, opened 2023-04-04
Remove exec-bit on .viminspector.json

atombrella:
not really sure why this file had the exec-bit set, but it seems like a silly thing to have :D


simar7 (approved):
thanks lgtm

PR #1269, opened 2023-04-06
Fix: correct search for control plane containers

alex123012:
Now in functions `is_apiserver`, `is_etcd`, `is_controllermananager` and `is_scheduler` regexes for first command entry are not "strict", so when there is more than one container in control plane pod with first command entries, for example, `kube-apiserver` and `kube-apiserver-healthcheck` - checks for control plane containers would be applied also for another (non control plane) container. This produces false negative check results.

In this PR I proposed solution for fixing this behavior.

### DIFF
--- a/rules/kubernetes/lib/kubernetes.rego
+++ b/rules/kubernetes/lib/kubernetes.rego
@@ -179,17 +179,17 @@ command_has_flag(command, flag) {
 }
 
 is_controllermananager(container) {
-	regex.match("kube-controller-manager", container.command[0])
+	regex.match("^kube-controller-manager$", container.command[0])
 }
 
 is_etcd(container) {
-	regex.match("etcd", container.command[0])
+	regex.match("^etcd$", container.command[0])
 }
 
 is_scheduler(container) {
-	regex.match("kube-scheduler", container.command[0])
+	regex.match("^kube-scheduler$", container.command[0])
 }
 
 is_apiserver(container) {
-	regex.match("kube-apiserver", container.command[0])
+	regex.match("^kube-apiserver$", container.command[0])
 }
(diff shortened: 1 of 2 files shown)


simar7 (dismissed):
could you give an example of how this would fail today (false positive)?

simar7:
adding @chen-keinan to take a second look

alex123012:
@simar7 Hi!
So, we are trying to run CIS benchmark with trivy-operator and receive fail for all `kube-apiserver` checks. Our `kube-apiserver` pod looks like:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: kube-apiserver
  namespace: kube-system
spec:
  containers:
  - command:
    - kube-apiserver
    - "..."
    name: kube-apiserver
  - command:
    - /usr/local/bin/kube-apiserver-healthcheck
    name: healthcheck
```

With old regex it will check both containers in `kube-apiserver` pod:
image
(look at dev tools)

If we use regex from this PR - all works as expected (only `kube-apiserver` container is checked):
image

As you can see, examples differ only by regex in `is_apiserver` func

[playground](https://play.openpolicyagent.org/p/ajIpoTZ2US)

alex123012:
@simar7 And hi here:)
I've added tests not to cis benchmark rules, but for lib itself.
Also i've fixed regex to proper validate abs paths for control plane binaries in pod command (refer to tests)

simar7 (approved):

simar7:
> @simar7 And hi here:)
> I've added tests not to cis benchmark rules, but for lib itself.
> Also i've fixed regex to proper validate abs paths for control plane binaries in pod command (refer to tests)

awesome, thanks!

PR #1270, opened 2023-04-07
Check for `--no-install-recommends` in apt-get

atombrella:
The check has mostly been adapted from the existing `-y` check.

There's an alternative way to deal with this, but it may require some more work. You can use `APT::Install-Recommends=false` in `/etc/apt/apt.conf` (and as `-o "APT::Install-Recommends=false"`), but perhaps the `--no-install-recommends` is the most commonly used way of dealing with this.

Hadolint has https://github.com/hadolint/hadolint/blob/master/src/Hadolint/Rule/DL3015.hs as an equivalent check.


simar7 (approved):
awesome @atombrella - this lgtm, appreciate adding the thorough tests!

PR #1278, opened 2023-04-10
WIP: Fix Terraform VPC adapter flow logs detection

Octogonapus:
There is a bug with detecting if a VPC has flow logs enabled, which I have traced to the Terraform VPC adapter code.
I tried to fix this, but I'm new to the project and I got stuck at trying to recursively resolve the VPC ID.
I have created a failing test case which reproduces the bug.

The crux of the issue is that the adapter can't resolve the VPC ID through the local.
I would assume some sort of recursive resolver is required to go "though" the local to the VPC block.

Also FWIW this test case is based on the very popular [terraform-aws-vpc](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/60516929e7330f7929cb5c55c6e43fa2427803a8/main.tf#L20) module.

This bug does not occur when scanning a Terraform plan file; it occurs only when scanning the Terraform source code.


Octogonapus:
This is still a bug. I don't know how to fix it. Maybe a maintainer could comment with some guidance?

Octogonapus:
I have implemented a minimal solution, but it's quite hacky. I'm not sure what the proper way to do this would be. From my poking around, it seems like the parser just doesn't handle locals much?

Octogonapus:
Implemented via https://github.com/aquasecurity/defsec/pull/1422

PR #1285, opened 2023-04-17
feat: GCP compute instance rule 'enable secure boot'

aisha-als:
Adding GCP rule for enable secure boot within shielded VM:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#enable_secure_boot


simar7 (approved):

simar7:
thanks lgtm!

PR #1295, opened 2023-04-27
fix: kubelet policies --read-only-port and --anonymous-auth

chen-keinan:
Related https://github.com/aquasecurity/trivy-operator/issues/1181


chen-keinan:
@simar7 please review this PR when you have time

simar7 (approved):
lgtm but I will only be able to cut a release after this is merged: https://github.com/aquasecurity/defsec/pull/1302 as it is a critical fix that needs to happen in the next release.

PR #1153, opened 2023-01-27
added autoscaling adapter

realwebdev:

PR #1297, opened 2023-04-27
Revert "added autoscaling adapter (#1153)"

simar7:
This reverts commit 124ecc6d5b1ff5b1dec7dddb0fdf8fe620688d40.

Ref: https://github.com/aquasecurity/trivy/issues/4124


simar7:
@knqyf263 I left more details here https://github.com/aquasecurity/defsec/pull/1153#issuecomment-1526826274

The way to proceed here will be to update the bundle version.

PR #1302, opened 2023-05-03
fix(adapters): Revert new adapters

simar7:
The newly added adapters require a `MAJOR` policy bundle update so that they can work with older versions of Trivy. 

Without a new `MAJOR` policy bundle update, the following issue will happen: https://github.com/aquasecurity/trivy/issues/4124

The plan is to merge all adapters at once (for all new AWS services at least), when they are all ready. At that point we will bump the `MAJOR` policy bundle version as per short term proposal idea described in this proposal here: https://github.com/aquasecurity/trivy/issues/4134


simar7:
hi @knqyf263 - if you have time could you help me test this? Below are the instructions to test:

0. Run a local docker registry to push the bundle to:
```bash
docker run -it --rm -p 5000:5000 registry
```

1. Create a policy bundle with this branch. You can use the following operations:
```bash
rm -rf bundle bundle.tar.gz && make bundle  && oras push localhost:5000/defsec:1 --config /dev/null:application/vnd.cncf.openpolicyagent.config.v1+json  bundle.tar.gz:application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip
```

2. Build an "old" version of Trivy (anything works, I selected the v0.40.0 tag), build it with the following patch:
```patch
diff --git a/pkg/policy/policy.go b/pkg/policy/policy.go
index 19d27bd34..1e4c3076b 100644
--- a/pkg/policy/policy.go
+++ b/pkg/policy/policy.go
@@ -18,8 +18,10 @@ import (
 )
 
 const (
-       bundleVersion    = 0 // Latest released MAJOR version for defsec
-       bundleRepository = "ghcr.io/aquasecurity/defsec"
+       bundleVersion    = 1 // Latest released MAJOR version for defsec
+       bundleRepository = "localhost:5000/defsec"
        policyMediaType  = "application/vnd.cncf.openpolicyagent.layer.v1.tar+gzip"
        updateInterval   = 24 * time.Hour
 )
```

Build with `mage build` as normal after applying this patch.

3. Run Trivy as following:
```bash
rm -rf ~/Library/Caches/trivy/ &&  ~/repos/trivy/trivy --debug config .
```

Where the `trivy` is the locally built binary in step 2. The above should clear the policy  bundle on disk, download a locally built one from step 1, complete the scan and not throw any errors.

knqyf263:
I confirmed it worked.

```
$ rm -rf ~/Library/Caches/trivy/policy/content/policies
$ make bundle
$ tar xvf bundle.tar.gz -C ~/Library/Caches/trivy/policy/content 
$ trivy config .
```

PR #1309, opened 2023-05-06
fix: Revert AVD-AWS-0342 policy

simar7:
This policy seems to be triggered on any input. Even when no relevant resources are present, will fail this policy.


simar7:
@giorod3 can you review this as you had merged https://github.com/aquasecurity/defsec/pull/1259 so I am hoping you have some more context here how this policy works.

simar7:
This PR is currently blocking https://github.com/aquasecurity/trivy/pull/4178

giorod3 (approved):

PR #1311, opened 2023-05-07
bug fix : misspelling

testwill:


simar7:
could you update here as well https://github.com/aquasecurity/defsec/blob/673ff3afe3745a8ee229078e53ccae0250964c46/rules/kubernetes/lib/kubernetes.rego#L143-L150

simar7 (approved):
lgtm, thanks!

PR #1324, opened 2023-05-23
Support checking for the last user as 'root' with a group specified, …

tspearconquest:
…and checking for a root user specified as UID 0, with or without a group name or GID number specified


tspearconquest:
I found the steps in the Makefile and was able to verify the tests fail against master but pass here. :)

simar7 (approved):

PR #1326, opened 2023-05-23
Improve the command splitting

tspearconquest:
This matches the split used in `run_command_cd_instead_of_workdir.rego`


simar7:
is there a case where the existing implementation fails?

tspearconquest:
I can't think of any off the top of my head other than if someone has a typo like ls /usr/bin&& ls /root where the space is missing before or after the && which would cause a failure to build the image anyway since it's not proper shell syntax.

That said, the old way does not match the spaces and could result in the spaces finding their way into commands_list, which probably isn't intended.

simar7 (approved):

PR #1334, opened 2023-05-30
added logic to pass config data settings to rego rules

AkhtarAmir:


Makefile
```diff
@@ -76,3 +76,7 @@ outdated-api-updated:
 .PHONY: bundle
 bundle:
 	./scripts/bundle.sh
+
+.PHONY: build
+build:
+	go build -o bin/defsec ./cmd/defsec
```
AkhtarAmir, line 82:
```suggestion
```

pkg/scanners/cloud/aws/options.go
```diff
@@ -31,6 +31,14 @@ func ScannerWithAWSRegion(region string) options.ScannerOption {
...
 }
 
+func ScannerWithRegoOnly(regoOnly bool) options.ScannerOption {
+	return func(s options.ConfigurableScanner) {
+		if aws, ok := s.(ConfigurableAWSScanner); ok {
+			aws.SetRegoOnly(regoOnly)
+		}
+	}
+}
+
```
AkhtarAmir, line 41:
```suggestion
```

simar7:
Could you add some unit tests that exercise this logic? You can see some here for example on how to write one https://github.com/aquasecurity/defsec/blob/master/pkg/scanners/cloud/aws/scanner_test.go

pkg/scanners/cloud/aws/scanner_test.go
```diff
@@ -62,6 +62,7 @@ func Test_AWSInputSelectors(t *testing.T) {
 	testCases := []struct {
 		name            string
 		srcFS           fs.FS
+		dataFS          fs.FS
```
simar7, line 65: I see you've added this field but haven't defined any cases in this test. I think we should move the test you added as a test case here as most of the logic is common between the two.

pkg/scanners/cloud/aws/scanner_test.go
```diff
@@ -437,3 +439,98 @@ deny[res] {
...
+		srcFS           fs.FS
+		dataFS          fs.FS
+		state           state.State
+		expectedResults struct {
+			totalResults int
+			summaries    []string
+		}
+	}{
+		{
+			name: "single cloud, single selector with config data",
```
simar7, line 455: This is good. We can add another case where the value is a different type, for instance an int.

simar7 (changes requested):
Good start, left some comments.

pkg/scanners/cloud/aws/scanner_test.go
AkhtarAmir: We are adding to these options in scanner
options.ScannerWithDataFilesystem(tc.dataFS),
options.ScannerWithDataDirs("config-data/")

so we will have to include config data this in all tests

pkg/scanners/cloud/aws/scanner_test.go
AkhtarAmir: okay

pkg/scanners/cloud/aws/scanner_test.go
simar7: You can make the options on the test cases itself rather than declared once.

simar7: Any update?

simar7:
@AkhtarAmir can you update this PR soon? I'm looking to cut a release end of the month (regular cadence for trivy and defsec) and would like to have this included.

PR #1339, opened 2023-06-13
support import block for tf 1.5.0+

nishigori:
docs:
- https://github.com/hashicorp/terraform/blob/v1.5/CHANGELOG.md#150-june-12-2023
- https://developer.hashicorp.com/terraform/language/import

New `import` block was added from Terraform 1.5.0
I added type support in a form similar to `moved` block.

### DIFF
--- a/pkg/scanners/terraform/parser/parser_test.go
+++ b/pkg/scanners/terraform/parser/parser_test.go
@@ -38,6 +38,10 @@ moved {
 
 }
 
+import {
+
+}
+
 resource "cats_cat" "mittens" {
 	name = "mittens"
 	special = true
@@ -45,7 +49,7 @@ resource "cats_cat" "mittens" {
 
 resource "cats_kitten" "the-great-destroyer" {
 	name = "the great destroyer"
-    parent = cats_cat.mittens.name
+	parent = cats_cat.mittens.name
 }
 
 data "cats_cat" "the-cats-mother" {
--- a/pkg/scanners/terraform/parser/parser_test.go
+++ b/pkg/scanners/terraform/parser/parser_test.go
@@ -39,7 +39,8 @@ moved {
 }
 
 import {
-
+  to = cats_cat.mittens
+  id = "mittens"
 }
 
 resource "cats_cat" "mittens" {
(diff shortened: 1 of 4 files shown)


pkg/scanners/terraform/parser/parser_test.go
```diff
@@ -38,14 +38,18 @@ moved {
 
 }
 
+import {
+
+}
+
```
simar7, line 44: I see that you've added this but none of the test assertions were modified. How can we test this change?

nishigori: Without the implementation code, the test will fail.
Here, as with the moved block, this means that the import block is not a syntax error.

Like:

```
# Case: Only changed test code
$ git diff origin/master
diff --git a/pkg/scanners/terraform/parser/parser_test.go b/pkg/scanners/terraform/parser/parser_test.go
index 8f693a0d..ebef5458 100644
--- a/pkg/scanners/terraform/parser/parser_test.go
+++ b/pkg/scanners/terraform/parser/parser_test.go
@@ -38,6 +38,10 @@ moved {

 }

+import {
+
+}
+
 resource "cats_cat" "mittens" {
        name = "mittens"
        special = true

$ make test
...
--- FAIL: Test_BasicParsing (0.00s)
    parser_test.go:68: test.tf:19,1-7: Unsupported block type; Blocks of type "import" are not expected here.
```

nishigori: @simar7 Do you mean you want more tests? (For example, for attributes in an import block)

simar7: > @simar7 Do you mean you want more tests? (For example, for attributes in an import block)

Yes that's right. I meant that we should test this import block functionality by having attributes in the import block.

simar7:
Just reading the release notes and it seems this feature isn't fully ready just yet?

> This is an early version of the import block feature, for which we are actively seeking user feedback to shape future development. The import block currently does not support interpolation in the id field, which must be a string.


If that's the case, I propose we should wait until they have finalized the spec for it.

pkg/scanners/terraform/parser/parser_test.go
nishigori: Added there testing from 4ca5e4f

nishigori:
@simar7 Yes, it seems this feature isn't fully ready just yet.
My suggestion is to have defsec support just to say that the import block exists.
This is because tfsec is giving me an error and I want to suppress it ( https://github.com/aquasecurity/tfsec/issues/2070 )

simar7:
> @simar7 Yes, it seems this feature isn't fully ready just yet.
> My suggestion is to have defsec support just to say that the import block exists.
> This is because tfsec is giving me an error and I want to suppress it ( [aquasecurity/tfsec#2070](https://github.com/aquasecurity/tfsec/issues/2070) )

I see - that's a great point. Your suggestion makes sense to include this in. Thanks for the PR!

simar7:
Associated issue for tracking https://github.com/aquasecurity/trivy/issues/4657

simar7 (approved):

PR #1340, opened 2023-06-14
fix(k8s): Ignore case when evaluating KSV106

simar7:
Fixes: https://github.com/aquasecurity/trivy/discussions/4366

Signed-off-by: Simar <simar@linux.com


chen-keinan:
lgtm

chen-keinan (approved):

joebowbeer:
@simar7 this is not strictly correct, and in practice it may not catch an issue that will be rejected by a compliant admission controller in the cluster: negating one of the benefits of using trivy in CI/CD.

PR #1361, opened 2023-07-01
Support define api-versions for helm scanner

jkroepke:
This PR add an function is pass `--api-versions` to the helm command.

This resolves an issue, where trivy reports a misconfiguration because of deprecated APIVersions, but the helm chart is using `Capabilities.APIVersions.Has` to validate if the newer API Version is available. 

```yaml
apiVersion: {{ $.Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" | ternary "policy/v1" "policy/v1beta1" }}
kind: PodDisruptionBudget
```

`Capabilities.APIVersions.Has` does not work well, since defsec is running helm with `--dry-run --client`

### DIFF
--- a/pkg/scanners/helm/options.go
+++ b/pkg/scanners/helm/options.go
@@ -41,3 +41,11 @@ func ScannerWithStringValues(values ...string) options.ScannerOption {
 		}
 	}
 }
+
+func ScannerWithAPIVersions(values ...string) options.ScannerOption {
+	return func(s options.ConfigurableScanner) {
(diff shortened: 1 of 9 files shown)


simar7:
thanks for the PR - do you have an associated discussion to describe what are you trying to achieve? If not, you can file one here: https://github.com/aquasecurity/trivy/discussions/new/choose

jkroepke:
https://github.com/aquasecurity/defsec/discussions/1372

simar7:
@chen-keinan - could you take a look?

chen-keinan (approved):

PR #1367, opened 2023-07-03
docs: fix the path to `policies`

nikpivkin:
Fixed paths to policies in the doc


simar7 (approved):
thanks for going through the documentation!

PR #1370, opened 2023-07-05
fix(terraform): allow nullable value for default values of vars

nikpivkin:
Evaluating an expression like `attr = var.some_var != null ? var.some_var : "default"` fails: `The true and false result expression must have consistent types`.
Terraform [supports null](https://developer.hashicorp.com/terraform/language/expressions/types#null) values that have no type. But defsec converts null with type [DynamicPseudoTypes](https://github.com/zclconf/go-cty/blob/7dcbae46a6f247e983efb1fa774d2bb68781a333/cty/unknown.go#L59) to `city.NilVal`, which cannot be used in conditional expressions.

See https://github.com/aquasecurity/trivy/discussions/4736

### DIFF
--- a/pkg/terraform/attribute.go
+++ b/pkg/terraform/attribute.go
@@ -253,6 +253,23 @@ func (a *Attribute) Value() (ctyVal cty.Value) {
 	return ctyVal
 }
 
+// Allows a null value for a variable https://developer.hashicorp.com/terraform/language/expressions/types#null
+func (a *Attribute) NullableValue() (ctyVal cty.Value) {
+	if a == nil {
+		return cty.NilVal
+	}
+	defer func() {
+		if err := recover(); err != nil {
+			ctyVal = cty.NilVal
+		}
+	}()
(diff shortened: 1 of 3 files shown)


pkg/terraform/attribute.go
```diff
@@ -253,6 +253,23 @@ func (a *Attribute) Value() (ctyVal cty.Value) {
...
+func (a *Attribute) NullableValue() (ctyVal cty.Value) {
+	if a == nil {
+		return cty.NilVal
+	}
+	defer func() {
+		if err := recover(); err != nil {
+			ctyVal = cty.NilVal
+		}
+	}()
+	ctyVal, _ = a.hclAttribute.Expr.Value(a.ctx.Inner())
```
simar7, line 266: what if `err` is not nil? what would `ctyVal` be in that case and would it still be OK to proceed ahead in the logic?

simar7 (approved):
lgtm

PR #1374, opened 2023-07-07
feat(terraform): support for multiple instances of the same resource

nikpivkin:
Now the resource instance overwrites the context variables of the previous instances, which makes it impossible to use multiple instances of the same resource.

See https://github.com/aquasecurity/trivy/issues/4627


simar7 (approved):
nice, lgtm!

PR #1382, opened 2023-07-10
Our objective is to align the titles and logic with the commercial version.

mjshastha:
Our objective is to align the titles and logic with the commercial version.

### DIFF
--- a/rules/kubernetes/policies/general/allowing_to_update_a_malicious_pod.rego
+++ b/rules/kubernetes/policies/general/allowing_to_update_a_malicious_pod.rego
@@ -1,5 +1,5 @@
 # METADATA
-# title: "Do not allow update/create of a malicious pod"
+# title: "Manage Kubernetes workloads and pods"
 # description: "Check whether role permits update/create of a malicious pod"
 # scope: package
 # schemas:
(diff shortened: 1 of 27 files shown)


simar7:
hi @chen-keinan could you take a look?

dheerajkadri:
Please cross check if the title updates are in sync with the AVD_DOCS

mjshastha:
> Please cross check if the title updates are in sync with the AVD_DOCS

verified

deven0t (approved):

rules/kubernetes/policies/general/allowing_to_update_a_malicious_pod.rego
```diff
@@ -20,9 +20,9 @@ package builtin.kubernetes.KSV048
 import data.lib.kubernetes
 import data.lib.utils
 
-workloads := ["deployments", "daemonsets", "statefulsets", "replicationcontrollers", "replicasets", "jobs", "cronjobs"]
+workloads := ["pods", "deployments", "jobs", "cronjobs", "statefulsets", "daemonsets", "replicasets", "replicationcontrollers"]
 
-changeVerbs := ["update", "create", "*"]
+changeVerbs := ["create", "update", "patch", "delete", "deletecollection", "impersonate", "*"]
```
dheerajkadri, line 25: wondering why we are using asterisk (*) here - wildcard character that matches any string. We have added more verbs now. Do we still need * ?

dheerajkadri: If we are using the wildcard (*), then there is no need to define verbs explicitly, since the asterisk(*) will match any verb.

mjshastha: I concur that including a wildcard (*) in the verbs list negates the need to define verbs explicitly, as the asterisk (*) will match any verb. The asterisk (*) was already present in the verbs list, and I introduced additional verbs to ensure alignment between the oss and commercial.

chen-keinan:
in general lgtm , see failing tests

mjshastha:
@chen-keinan I have addressed the issues with the failing tests.

chen-keinan:
@mjshastha could you please rebase your branch with upstream

simar7 (approved):

PR #1386, opened 2023-07-14
Update rego title in OSS defsec Part2

mjshastha:
we introduced three new rego checks to the oss. These checks are currently part of the commercial but not included in the oss.
Additionally, there are certain checks in the oss where two checks are combined into a single check. However, in the commercial, these checks are treated as two separate checks. Therefore, we will be splitting the combined checks from the oss into two distinct checks to align with the commercial implementation.
 
Resolves: SLK-69409


simar7:
@chen-keinan PTAL after you are back, thanks.

dheerajkadri (approved):
LGTM!!

chen-keinan:
lgtm

chen-keinan (approved):

simar7:
Looks like tests are failing and there's a merge conflict with upstream after the first PR was merged.

Issue #1024, opened 2022-10-20
bug: Azure KeyVault key expiration date validation fails when using time_offset.rfc3339


brentmannering:
**Describe the bug**
Running tfsec against a `azurerm_key_vault_key` fails when the expiration date is set to the `rfc3339` property of the `time_offset` resource.

```
panic: not a string

goroutine 1 [running]:

github.com/zclconf/go-cty/cty.Value.AsString({{{0x0?, 0x0?}}, {0x0?, 0x0?}})
	/home/runner/go/pkg/mod/github.com/zclconf/go-cty@v1.10.0/cty/value_ops.go:1256 +0x138
github.com/aquasecurity/defsec/internal/adapters/terraform/azure/keyvault.adaptKey(0xc00012db00)
	/home/runner/go/pkg/mod/github.com/aquasecurity/defsec@v0.76.0/internal/adapters/terraform/azure/keyvault/adapt.go:154 +0x1ca
```

**To Reproduce**
1. Create dummy terraform definition using example code below
2. Run tfsec scan across terraform definition
3. See error

**Expected behavior**
Using the `rfc3339` is correctly interpreted as a string value and the tfsec scan/validation passes successfully

**Output of your tfsec command with --debug flag**
```
Running tfsec to update results
command: --force-all-dirs,--exclude-downloaded-modules,--verbose,--soft-fail,--format=json,--out=...
07:57.983818100 system.info                      APP       tfsec.exe
07:57.983818100 system.info                      VERSION   v1.28.0
07:57.983818100 system.info                      OS        windows
07:57.983818100 system.info                      ARCH      amd64
07:57.983818100 system.info                      KERNEL    
07:57.983818100 system.info                      TERM      
07:57.983818100 system.info                      SHELL     
07:57.983818100 system.info                      GOVERSION go1.18.5
07:57.983818100 system.info                      GOROOT    /opt/hostedtoolcache/go/1.18.5/x64
07:57.983818100 system.info                      CGO       false
07:57.983818100 system.info                      CPUCOUNT  8
07:57.983818100 system.info                      MAXPROCS  8
07:57.983818100 system.info                      WORKDIR   C:\Users\...\AppData\Local\Programs\Microsoft VS Code
07:57.983818100 system.info                      UID       -1
07:57.983818100 system.info                      EUID      -1
07:57.983818100 system.info                      DOCKER    false
07:57.983818100 system.info                      CI        false
07:57.983818100 system.info                      HOSTNAME  ...
[...]

PR #1387, opened 2023-07-14
fix(terraform): check that the expiration_date is string

nikpivkin:
Fixes https://github.com/aquasecurity/defsec/issues/1024


simar7 (approved):
lgtm!

PR #1395, opened 2023-07-18
fix(cloudformation): fix panic when use pseudo-parameters NoValue or NotificationARNs

nikpivkin:
Some pseudo-parameters are not strings, so when a parameter [is cast to a string](https://github.com/aquasecurity/defsec/blob/master/pkg/scanners/cloudformation/parser/fn_ref.go#L19), panic occurs. Also, the values of the `AWS::NotificationARNs` parameter must be properties, not strings, to be able to use with `Select` or `Join` functions.


simar7:
do we have an issue in trivy/defsec which this fixes?

simar7 (approved):

nikpivkin:
@simar7 I opened an [issue](https://github.com/aquasecurity/trivy/issues/4843)

simar7 (approved):

PR #1396, opened 2023-07-18
fix(cloudformation): resolve property depending on conditions

nikpivkin:
When a property is a value returned by a condition, then defsec does not evaluate it

Example:
```yaml
Resources:
  Bucket:
    Type: "AWS::S3::Bucket"
    Properties:
      VersioningConfiguration:
        !If [true, { Status: Enabled }, { Status: Suspended }]
```

Output:
```bash
....
AVD-AWS-0090 aws-s3-enable-versioning test.yaml:2-6
....
```

See https://github.com/aquasecurity/trivy/issues/4844

### DIFF
--- a/pkg/scanners/cloudformation/parser/property.go
+++ b/pkg/scanners/cloudformation/parser/property.go
@@ -241,13 +241,17 @@ func (p *Property) GetProperty(path string) *Property {
 	pathParts := strings.Split(path, ".")
 
 	first := pathParts[0]
-	var property *Property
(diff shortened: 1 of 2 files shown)


simar7:
do we have an issue in trivy/defsec which this fixes?

simar7 (approved):

nikpivkin:
@simar7 I opened an [issue](https://github.com/aquasecurity/trivy/issues/4844)

PR #1402, opened 2023-07-25
fix(azure): `SecurityAlertPolicies`: don't warn if policy is disabled

Frankkkkk:
Hi,

Azure's MSSql SecurityAlertPolicies can be disabled. In this case, it's okay to not have email addresses.

Cheers

### DIFF
--- a/internal/adapters/arm/database/mssql.go
+++ b/internal/adapters/arm/database/mssql.go
@@ -44,6 +44,7 @@ func adaptSecurityAlertPolicies(resource azure.Resource, deployment azure.Deploy
 	for _, policy := range deployment.GetResourcesByType("Microsoft.Sql/servers/securityAlertPolicies") {
 		policies = append(policies, database.SecurityAlertPolicy{
 			Metadata:           policy.Metadata,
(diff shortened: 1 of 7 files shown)


internal/adapters/arm/database/mssql.go
```diff
@@ -44,6 +44,7 @@ func adaptSecurityAlertPolicies(resource azure.Resource, deployment azure.Deploy
 	for _, policy := range deployment.GetResourcesByType("Microsoft.Sql/servers/securityAlertPolicies") {
 		policies = append(policies, database.SecurityAlertPolicy{
 			Metadata:           policy.Metadata,
+			State:              resource.Properties.GetMapValue("state").AsStringValue("", resource.Metadata),
```
simar7, line 47: if `resource.Properties.GetMapValue("state")` is nil, this will panic. We should check the state first and then use it.

simar7 (changes requested):
left comment

internal/adapters/arm/database/mssql.go
Frankkkkk: Hi @simar7 , TBH I don't know how to make it "right". Do you have any tips ? Thanks

PR #1405, opened 2023-07-27
fix: set HostnameImmutable if a custom endpoint is used

nikpivkin:
Amazon S3 supports both virtual-hosted–style and path-style URL acces, but the `GetBucketLocation` method uses virtual-hosted–style, which does not support localstack. `HostnameImmutable` flag [does not allow](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/endpoints.go#L102) the sdk to change the endpoint. This will allow us to use localstack for local testing.

See https://github.com/aquasecurity/trivy/issues/4876

### DIFF
--- a/internal/adapters/cloud/aws/adapt.go
+++ b/internal/adapters/cloud/aws/adapt.go
@@ -112,9 +112,10 @@ type resolver struct {
 
 func (r *resolver) ResolveEndpoint(_, _ string, _ ...interface{}) (aws.Endpoint, error) {
 	return aws.Endpoint{
-		URL:           r.endpoint,
-		SigningRegion: "custom-signing-region",
-		Source:        aws.EndpointSourceCustom,
+		URL:               r.endpoint,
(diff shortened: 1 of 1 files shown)


nikpivkin:
@simar7 I found information that amazon [has a S3 path deprecation plan,](https://aws.amazon.com/blogs/storage/update-to-amazon-s3-path-deprecation-plan/) but there is no exact date. Therefore, does it make sense to add this?
Alternatively, we can use a [network alias](https://docs.docker.com/engine/reference/commandline/network_connect/) for local testing.

simar7:
> @simar7 I found information that amazon [has a S3 path deprecation plan,](https://aws.amazon.com/blogs/storage/update-to-amazon-s3-path-deprecation-plan/) but there is no exact date. Therefore, does it make sense to add this?

I see. Yeah we shouldn't add this if it is planned to be deprecated.

PR #1423, opened 2023-08-10
fix(aws): remove duplicate bucket logging rule

nikpivkin:
We [already have](https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/s3/enable_bucket_logging.go) a bucket logging rule.

See https://github.com/aquasecurity/trivy/issues/4975
Fixes https://github.com/aquasecurity/trivy/issues/5004


nikpivkin:
@simar7 The tests failed at the OPA install stage: `sha256sum: checksum: no properly formatted SHA256 checksum lines found`

simar7:
These are the rules I moved into the `rules/` dir recently https://github.com/aquasecurity/defsec/commit/91898b5a97b9f7952205e8c2544ebcc9b3602fae do we have any other rules that also exist in Go that overlap with these? If so, we should remove them (Go rules).

nikpivkin:
@simar7 OK, I'll return rego and remove the GO rule. By the way, the rego rule contained an invalid AVD ID.

davidbmleach:
Is it possible this duplication also generates a false positive when you do something like this?:
```
# tfsec:ignore:aws-s3-enable-bucket-logging
# tfsec:ignore:aws-s3-enable-versioning
resource "aws_s3_bucket" "this" {
  bucket        = var.name
}
```

nikpivkin:
Hi @davidbmleach 

If [this comment](https://github.com/aquasecurity/defsec/issues/1427#issuecomment-1676669613) doesn't help you, could you describe in more detail what happened to you and what you expected to see?

rules/cloud/policies/aws/s3/enable_bucket_logging.cf.go
```diff
@@ -1,44 +0,0 @@
-package s3
```
simar7, line 1: We need to keep the examples (in Go) still available to us as they are used for doc generation. An example on how to use them is here https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/rds/disable_public_access.rego#L22-L25

rules/cloud/policies/aws/s3/enable_bucket_logging.tf.go
```diff
@@ -1,37 +0,0 @@
-package s3
```
simar7, line 1: ditto for terraform example.

simar7 (changes requested):
left comments

rules/cloud/policies/aws/s3/enable_bucket_logging.cf.go
nikpivkin: @simar7 The `good_example` field specifies the GO file, but the file also contains a bad example and links. How does it work?

simar7: @nikpivkin they are loaded like such https://github.com/aquasecurity/defsec/blob/063dcd6ec8d061cae23ff208cf7a71c6eed6a8fc/pkg/rego/metadata.go#L266

nikpivkin: @simar7 I think I [found](https://github.com/aquasecurity/defsec/blob/master/cmd/avd_generator/main.go#L88) a code that parses GO files to generate documentation.

I have a question:
- Why are only good examples generated for documentation, but not bad ones?

nikpivkin: @simar7 By the way, the "RDS Publicly Accessible" rule is duplicated:
- rego https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/rds/disable_public_access.rego
- GO https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/rds/no_public_db_access.go

nikpivkin:
@simar7 I updated the config for `golangci-lint` and added ignoring `U1000` (unused variables) for variables in those files referenced from Rego. I noticed that now `golangci-linter` uses the following linters to check dead code: `unused`, `struct check`, `deadcode` and `varcheck`. All of them except `unsuded` are [deprecated](https://github.com/golangci/golangci-lint/issues/1841). How about updating the config?

simar7:
> @simar7 I updated the config for `golangci-lint` and added ignoring `U1000` (unused variables) for variables in those files referenced from Rego. I noticed that now `golangci-linter` uses the following linters to check dead code: `unused`, `struct check`, `deadcode` and `varcheck`. All of them except `unsuded` are [deprecated](https://github.com/golangci/golangci-lint/issues/1841). How about updating the config?

Sounds good. Let's create a separate issue and track it there.

rules/cloud/policies/aws/s3/enable_bucket_logging.cf.go
simar7: > @simar7 By the way, the "RDS Publicly Accessible" rule is duplicated:
> 
> * rego https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/rds/disable_public_access.rego
> * GO https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/rds/no_public_db_access.go

New issue https://github.com/aquasecurity/trivy/issues/5065

simar7 (approved):

PR #1425, opened 2023-08-12
fix: make operations with reference key safe

nikpivkin:
See https://github.com/aquasecurity/trivy/issues/4983

### DIFF
--- a/pkg/terraform/reference.go
+++ b/pkg/terraform/reference.go
@@ -114,8 +114,7 @@ func (r Reference) String() string {
 	return base
 }
 
-func (r Reference) RefersTo(a Reference) bool {
-	other := a
+func (r Reference) RefersTo(other Reference) bool {
 
 	if r.BlockType() != other.BlockType() {
 		return false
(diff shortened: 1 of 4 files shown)


pkg/terraform/reference.go
```diff
@@ -133,33 +132,43 @@ func (r Reference) RefersTo(a Reference) bool {
 }
 
 func (r *Reference) SetKey(key cty.Value) {
+	if key.IsNull() || !key.IsKnown() {
```
simar7, line 135: Wouldn't this panic if key was actually null? In that case the `!key.IsKnown()` would panic.

simar7: ping @nikpivkin

nikpivkin: @simar7 The right operand of a boolean expression is evaluated [conditionally](https://go.dev/ref/spec#Logical_operators). If the `key` is null, then the left operand of the expression in the condition will be true and the function will exit, without checking whether the `key` is known.  And in any case, why should this check cause panic?

simar7: I see. I didn’t know the conditional evaluation of Go for such a case. This seems to follow the short circuit behaviour. 

Thanks this works fine then.

simar7 (approved):

PR #1439, opened 2023-08-30
ci: make the OPA installation more reliable

nikpivkin:
I also took out the installation of OPA in a composite action

Close https://github.com/aquasecurity/defsec/issues/1440

### DIFF
--- a/.github/actions/setup-opa/action.yaml
+++ b/.github/actions/setup-opa/action.yaml
@@ -7,7 +7,7 @@ runs:
       shell: bash
       run: |
         curl --retry 3 -L -o opa_linux_amd64 https://github.com/open-policy-agent/opa/releases/latest/download/opa_linux_amd64_static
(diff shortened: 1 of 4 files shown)


.github/actions/setup-opa/action.yaml
```diff
@@ -0,0 +1,13 @@
+name: Setup OPA CLI
+description: Setup OPA CLI
+runs:
+  using: composite
```
simar7, line 4: Nice! besides refactoring into a separate YAML, what other advantages does it have?

simar7 (approved):

.github/actions/setup-opa/action.yaml
nikpivkin: Added the retry flag to repeat the operation after a failure and changed the downloading of the OPA from the site to GitHub (I think GitHub is a more reliable source)

PR #1443, opened 2023-08-31
feat(cloud): add the DeletionProtection attribute to the RDS Cluster

dorayakikun:
Fixes https://github.com/aquasecurity/trivy/issues/5112

## motivation

I want to avert some human mistakes in the RDS Cluster, but I can't detect them now.
( Now, I can only detect the RDS instance's deletionProtection. )

For example, the below codes are not working as intended.

```
deny[res] {
	cluster := input.aws.rds.clusters[_]
        instance := cluster.instances[_]
	not instance.deletionprotection.value
	res := result.new("Cluster does not have Deletion Protection enabled", instance.deletionprotection)
}
```

So, I would like to add the DeletionProtection attribute to the RDS Cluster.

### DIFF
--- a/rules/cloud/policies/aws/rds/enable_cluster_deletion_protection.rego
+++ b/rules/cloud/policies/aws/rds/enable_cluster_deletion_protection.rego
@@ -1,6 +1,6 @@
 # METADATA
-# title: "RDS Deletion Protection Disabled"
-# description: "Ensure deletion protection is enabled for RDS database instances."
+# title: "RDS Cluster Deletion Protection Disabled"
+# description: "Ensure deletion protection is enabled for RDS clusters."
 # scope: package
 # schemas:
 # - input: schema["cloud"]
--- a/rules/cloud/policies/aws/rds/enable_cluster_deletion_protection.rego
+++ b/rules/cloud/policies/aws/rds/enable_cluster_deletion_protection.rego
@@ -1,0 +1,28 @@
+# METADATA
+# title: "RDS Deletion Protection Disabled"
+# description: "Ensure deletion protection is enabled for RDS database instances."
+# scope: package
+# schemas:
+# - input: schema["cloud"]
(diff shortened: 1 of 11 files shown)


simar7:
> ## motivation
> I want to avert some human mistakes in the RDS Cluster, but I can't detect them now. ( Now, I can only detect the RDS instance's deletionProtection. )
> 
> For example, the below codes are not working as intended.
> 
> ```
> deny[res] {
> 	cluster := input.aws.rds.clusters[_]
>         instance := cluster.instances[_]
> 	not instance.deletionprotection.value
> 	res := result.new("Cluster does not have Deletion Protection enabled", instance.deletionprotection)
> }
> ```
> 
> So, I would like to add the DeletionProtection attribute to the RDS Cluster.

I see. Thanks for the PR. Would you also mind creating a new Rego Rule for the clusters too? It would be similar to this rule we have for instances: https://github.com/aquasecurity/defsec/blob/master/rules/cloud/policies/aws/rds/enable_deletion_protection.rego - I think it could be beneficial for other users too.

simar7 (changes requested):
left a comment and you would also need to run `make schema` and commit the changes to pass the tests.

dorayakikun:
I appreciate your warm dealing.
I've tried to add docs and policies.

Are these processes correct?
If there are any problems, please let me know.

rules/cloud/policies/aws/rds/enable_cluster_deletion_protection.rego
```diff
@@ -0,0 +1,28 @@
+# METADATA
+# title: "RDS Deletion Protection Disabled"
+# description: "Ensure deletion protection is enabled for RDS database instances."
```
simar7, line 3: This should be changed to adapt for clusters instead of instances.

simar7 (approved):

PR #1449, opened 2023-09-07
feat: Add support for `skip_final_snapshot`

dorayakikun:
related to https://github.com/aquasecurity/trivy/issues/5147

### DIFF
--- a/internal/adapters/cloud/aws/rds/rds.go
+++ b/internal/adapters/cloud/aws/rds/rds.go
@@ -268,6 +268,7 @@ func (a *adapter) adaptCluster(dbCluster types.DBCluster) (*rds.Cluster, error) 
 		LatestRestorableTime: defsecTypes.TimeUnresolvable(dbClusterMetadata),
 		AvailabilityZones:    availabilityZones,
 		DeletionProtection:   defsecTypes.Bool(aws.ToBool(dbCluster.DeletionProtection), dbClusterMetadata),
+		SkipFinalSnapshot:    defsecTypes.Bool(false, dbClusterMetadata),
 	}
 
 	return cluster, nil
--- a/internal/adapters/cloudformation/aws/rds/cluster.go
+++ b/internal/adapters/cloudformation/aws/rds/cluster.go
@@ -28,6 +28,7 @@ func getClusters(ctx parser.FileContext) (clusters map[string]rds.Cluster) {
 			Engine:               defsecTypes.StringDefault(rds.EngineAurora, clusterResource.Metadata()),
 			LatestRestorableTime: defsecTypes.TimeUnresolvable(clusterResource.Metadata()),
 			DeletionProtection:   defsecTypes.BoolDefault(false, clusterResource.Metadata()),
+			SkipFinalSnapshot:    defsecTypes.BoolDefault(false, clusterResource.Metadata()),
 		}
 
 		if engineProp := clusterResource.GetProperty("Engine"); engineProp.IsString() {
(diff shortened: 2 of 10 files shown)


internal/adapters/cloud/aws/rds/rds.go
```diff
@@ -268,6 +268,7 @@ func (a *adapter) adaptCluster(dbCluster types.DBCluster) (*rds.Cluster, error)
 		LatestRestorableTime: defsecTypes.TimeUnresolvable(dbClusterMetadata),
 		AvailabilityZones:    availabilityZones,
 		DeletionProtection:   defsecTypes.Bool(aws.ToBool(dbCluster.DeletionProtection), dbClusterMetadata),
+		SkipFinalSnapshot:    defsecTypes.Bool(false, dbClusterMetadata),
```
dorayakikun, line 271: I'm implementing this modification to support `adapter-lint` .
The direct assignment of `false` is because there is no `SkipFinalSnapshot` in [the definitio](https://github.com/aws/aws-sdk-go-v2/blob/a81a9f719ac616318333ee18ee257110d11c95ff/service/rds/types/types.go#L432)n here.

If this modification is not appropriate, this pull request will Close 😢

internal/adapters/cloudformation/aws/rds/cluster.go
```diff
@@ -28,6 +28,7 @@ func getClusters(ctx parser.FileContext) (clusters map[string]rds.Cluster) {
 			Engine:               defsecTypes.StringDefault(rds.EngineAurora, clusterResource.Metadata()),
 			LatestRestorableTime: defsecTypes.TimeUnresolvable(clusterResource.Metadata()),
 			DeletionProtection:   defsecTypes.BoolDefault(false, clusterResource.Metadata()),
+			SkipFinalSnapshot:    defsecTypes.BoolDefault(false, clusterResource.Metadata()),
```
dorayakikun, line 31: It's the same as https://github.com/aquasecurity/defsec/pull/1449/files#r1319740047 .

internal/adapters/cloud/aws/rds/rds.go
simar7: I see. That's reasonable, it is strange that the definition does not contain this field.

simar7:
This looks good, can we also add a new rego policy which should check for this? Similar to how you added this https://github.com/aquasecurity/defsec/pull/1443/files#diff-78526a6a886c490234313f8e4a71f570d25b5d287fb1e36f9beee328d59f5a3d

internal/adapters/cloud/aws/rds/rds.go
dorayakikun: > it is strange that the definition does not contain this field.

I assume that the AWS team can't implement this attribute because it is not a state.
`-skip-final-shanpshot` is a flag in the `delete-db-cluster` command.

https://docs.aws.amazon.com/cli/latest/reference/rds/delete-db-cluster.html#options

However, I can set SkipFinalSnapshot as a attribute in Terraform because they can call the `delete-db-cluster` command when calling terraform destroy.

rules/cloud/policies/aws/rds/disable_cluster_skip_final_snapshot.rego
```diff
@@ -0,0 +1,28 @@
...
+#   severity: MEDIUM
+#   short_code: disable-cluster-skip-final-snapshot
+#   recommended_action: "Modify the RDS clusters to disable skip final snapshot."
+#   input:
+#     selector:
+#     - type: cloud
+#       subtypes:
+#         - service: rds
+#           provider: aws
+package PACKAGE_NAME
```
simar7, line 22: update this to the correct package name

rules/cloud/policies/aws/rds/disable_cluster_skip_final_snapshot_test.rego
```diff
@@ -0,0 +1,11 @@
+package PACKAGE_NAME
```
simar7, line 1: ditto

simar7 (approved):

PR #1467, opened 2023-10-05
fix(terraform): do not scan local modules as root modules

nikpivkin:
This PR fixes a bug when Trivy scans local modules as root modules. See https://github.com/aquasecurity/trivy/issues/4988.

Configuration example:

`main.tf`
```tf
module "main" {
  source      = "./modules/s3"
  versioning = "Enabled"
}
```

`modules/s3/main.tf`
```tf
variable "versioning" {
  type = string
}

resource "aws_s3_bucket" "main" {
  bucket = "test"
}

resource "aws_s3_bucket_versioning" "example" {
  bucket = aws_s3_bucket.main.id
  versioning_configuration {
    status = var.versioning
  }
}
```

Before:
```bash
trivy config . -f json | jq '.Results[] | .Misconfigurations | .[]?.Title'
2023-10-05T20:10:23.028+0700	INFO	Misconfiguration scanning is enabled
2023-10-05T20:10:24.956+0700	INFO	Detected config files: 3
"S3 Access block should block public ACL"
"S3 Access block should block public ACL"
"S3 Access block should block public policy"
"S3 Access block should block public policy"
"Unencrypted S3 bucket."
"Unencrypted S3 bucket."
"S3 Bucket Logging"
"S3 Bucket Logging"
"S3 Data should be versioned"
"S3 Access Block should Ignore Public Acl"
"S3 Access Block should Ignore Public Acl"
"S3 Access block should restrict public bucket to limit access"
"S3 Access block should restrict public bucket to limit access"
"S3 buckets should each define an aws_s3_bucket_public_access_block"
"S3 buckets should each define an aws_s3_bucket_public_access_block"
"S3 encryption should use Customer Managed Keys"
"S3 encryption should use Customer Managed Keys"
```

After:
```bash
2023-10-05T20:13:02.416+0700    INFO    Misconfiguration scanning is enabled
2023-10-05T20:13:02.839+0700    INFO    Detected config files: 2
"S3 Access block should block public ACL"
"S3 Access block should block public policy"
"Unencrypted S3 bucket."
"S3 Bucket Logging"
"S3 Access Block should Ignore Public Acl"
"S3 Access block should restrict public bucket to limit access"
"S3 buckets should each define an aws_s3_bucket_public_access_block"
"S3 encryption should use Customer Managed Keys"
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -80,6 +80,7 @@ require (
 	github.com/docker/docker v24.0.5+incompatible
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/testcontainers/testcontainers-go v0.22.0
+	golang.org/x/exp v0.0.0-20230510235704-dd950f8aeaea
 	k8s.io/utils v0.0.0-20230220204549-a5ecb0141aa5
 )
 
@@ -238,7 +239,6 @@ require (
 	go.opentelemetry.io/otel v1.14.0 // indirect
 	go.opentelemetry.io/otel/trace v1.14.0 // indirect
 	go.starlark.net v0.0.0-20200306205701-8dd3e2ee1dd5 // indirect
-	golang.org/x/exp v0.0.0-20230510235704-dd950f8aeaea // indirect
 	golang.org/x/mod v0.10.0 // indirect
 	golang.org/x/net v0.10.0 // indirect
 	golang.org/x/oauth2 v0.7.0 // indirect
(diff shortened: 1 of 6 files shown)


nikpivkin:
@simar7 The test failed due to a bug in go1.19 on windows. https://github.com/golang/go/issues/51007

simar7:
I see. Should we update the version of Go in defsec prior to merging this PR or just ignore the error for now? 

If upgrading the version isn't too time consuming let's do that. What do you think? Since Trivy already uses 1.20.

nikpivkin:
@simar7 I'll do it.

simar7 (approved):

PR #1471, opened 2023-10-06
feat(terraform): add an option to use FS for configs

nikpivkin:
See https://github.com/aquasecurity/trivy/issues/5280

### DIFF
--- a/pkg/scanners/terraform/options.go
+++ b/pkg/scanners/terraform/options.go
@@ -1,6 +1,7 @@
 package terraform
 
 import (
+	"io/fs"
 	"strings"
 
 	"github.com/aquasecurity/defsec/pkg/scan"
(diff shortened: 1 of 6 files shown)


pkg/scanners/terraform/options.go
```diff
@@ -193,10 +193,10 @@ func ScannerWithDownloadsAllowed(allowed bool) options.ScannerOption {
 	}
 }
 
-func ScannerWithEmbeddedLibraries(embedded bool) options.ScannerOption {
```
nikpivkin, line 196: This option was not used because there is a global option for using embedded libraries.  https://github.com/aquasecurity/defsec/blob/master/pkg/scanners/options/scanner.go#L62

nikpivkin:
@simar7  I'll open this PR in [trivy-iac](https://github.com/aquasecurity/trivy-iac).

PR #1488, opened 2023-10-30
feat: generate allowed actions

nikpivkin:
Added a command to generate a file with actions that do not support resource level permissions. This will allow periodic synchronisation with AWS. There are about 5135 such actions now (vs 3600 in the main branch).

### Related issues
- https://github.com/aquasecurity/trivy/issues/5040


simar7:
How often does this list change? While the logic to auto generate is nice, I'm afraid it adds unnecessary complexity and is prone to breakage, if AWS decides to change their page which is likely as it isn't a contract/API.

nikpivkin:
@simar7 I found a repository with documentation, but it's already archived. I don't know how often this list changes, but so far more than 1k actions have been added (or changed) compared to the main branch. Is there any idea how we can update the actions?

nikpivkin:
@simar7 I found a [js](https://awsiamconsole.s3.amazonaws.com/iam/assets/js/bundles/policies.js) resource that is on a policy generation [site](http://awspolicygen.s3.amazonaws.com/policygen.html) that contains a json object with all the actions. I think this is a more reliable source compared to an html page. Wdyt?

nikpivkin:
[This repository](https://github.com/TryTryAgain/aws-iam-actions-list/commits/master/) contains a list of actions and it is updated frequently.

simar7:
> [This repository](https://github.com/TryTryAgain/aws-iam-actions-list/commits/master/) contains a list of actions and it is updated frequently.

Feels a little better than scraping, maybe we can go with this for the meantime since we don't have any better alternatives.

nikpivkin:
@simar7 I gave the example of this repository to show that changes to AWS services happen quite often. But this repository is not suitable for us because it only contains a list of actions, and we also need information about the resources required for each action.

simar7:
> @simar7 I gave the example of this repository to show that changes to AWS services happen quite often. But this repository is not suitable for us because it only contains a list of actions, and we also need information about the resources required for each action.

I guess it's fine as it is a self contained piece of code with no better alternatives at the moment.

nikpivkin:
@simar7 Test run of the workflow https://github.com/nikpivkin/defsec/commit/7c0ae8549c78e2e34afb867d3292637464aa38bf

There is also a problem with trivy-policies due to cyclic dependency. https://github.com/aquasecurity/defsec/actions/runs/7179365411

simar7:
> @simar7 Test run of the workflow [nikpivkin@7c0ae85](https://github.com/nikpivkin/defsec/commit/7c0ae8549c78e2e34afb867d3292637464aa38bf)

lgtm,  should we merge this PR?

simar7 (approved):

PR #1509, opened 2023-12-18
feat(ec2): name support for the launch template

nikpivkin:
See https://github.com/aquasecurity/trivy/issues/5793

### DIFF
--- a/pkg/terraform/presets.go
+++ b/pkg/terraform/presets.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/zclconf/go-cty/cty"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/terraform/presets.go
```diff
@@ -22,6 +23,9 @@ func createPresetValues(b *Block) map[string]cty.Value {
 	switch b.TypeLabel() {
 	case "aws_iam_policy_document":
 		presets["json"] = cty.StringVal(b.ID())
+	// If the user leaves the name blank, Terraform will automatically generate a unique name
+	case "aws_launch_template":
+		presets["name"] = cty.StringVal(uuid.New().String())
```
simar7, line 28: do we have to use a UUID here?

nikpivkin: We need to generate some unique value in case the user did not pass the `name` argument.

simar7 (approved):

PR #1544, opened 2025-01-21
docs: add migration notice

simar7:


Emyrk:
> @itaysk as a next step should we archive this repo, effectively making it read only?

Please do. Also archive https://github.com/aquasecurity/trivy-iac. I was unaware these two repos were stale.

itaysk (approved):
thanks

